PR #41, opened 2025-05-20
Remove interactive marking API

### LINKED ISSUE #20: Refactor the test `test_marking_operations`
This single test checks several things. I think this should be brokeninto separate tests with clearer names. Right it checks that:

+ an empty table returned by `get_markers` has the correct columns.
+ the `is_marking` attribute is not settable.
+ setting the marker style actually changes the marker properties
+ calling `start_marking` changes the properties `click_center`, `click_drag`, and `scroll_pan`
+ calling `stop_marking` returns those properties back the the values they had before `start_marking` was called, i.e. check that property cacheing is implemented.
+ no warning is issued when retrieving a marker set that does not exist
+  recheck that start/stop marking caches properly
+ checks that a simulated mouse click adds a marker **--> this is implementation is ginga- specific <--** so it should be removed

mwcraig:
This pull request removes the interactive marking API (`start_marking`, `stop_marking`, `is_marking`) following offline discussion.

Fixes #20 which was mostly about interactive marking.

### DIFF
--- a/src/astro_image_display_api/dummy_viewer.py
+++ b/src/astro_image_display_api/dummy_viewer.py
@@ -45,9 +45,6 @@ class ImageViewer:
     # Default marker name for marking via API
     DEFAULT_MARKER_NAME: str = ImageViewerInterface.DEFAULT_MARKER_NAME
 
-    # Default marker name for interactive marking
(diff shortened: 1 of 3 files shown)


pllim (approved):
Does simplify a lot of stuff. Thanks!

Now that we are not bound by interactive click to mark, we also have the option to define marker property as a dictionary mapping marker label name to its properties, as discussed offline yesterday, but that is a separate PR.

PR #42, opened 2025-05-20
Combine load_ methods into single load method

mwcraig:
This combines the `load_...` methods into a single `load` method.

Questions to resolve:

- [ ] Should the test class check for anything other than loading a FITS file?
- [ ] Should an image viewer have a method that indicates what types it can load?


src/astro_image_display_api/dummy_viewer.py
```diff
@@ -125,7 +125,7 @@ def cursor(self, value: str) -> None:
     # The methods, grouped loosely by purpose
 
     # Methods for loading data
-    def load_fits(self, file: str | os.PathLike) -> None:
+    def load(self, file: str | os.PathLike | ArrayLike | NDData) -> None:
```
pllim, line 128: For Roman, this could also be ASDF. I dunno how forgiving this typing stuff is.

pllim: Oh wait, this is the dummy viewer. Lemme move this comment.

src/astro_image_display_api/interface_definition.py
```diff
@@ -57,41 +57,19 @@ class ImageViewerInterface(Protocol):
 
     # The methods, grouped loosely by purpose
 
-    # Methods for loading data
+    # Method for loading data
     @abstractmethod
-    def load_fits(self, file: str | os.PathLike) -> None:
+    def load(self, data: Any) -> None:
```
pllim, line 62: What about the `**kwargs`? For example, FITS might need also the `ext` keyword to specific extension, but this keyword does not apply if one loads a pure numpy array, so we have to stay flexible here?

mwcraig: I'd like to add `**kwargs` everywhere at the same time...

src/astro_image_display_api/dummy_viewer.py
```diff
@@ -123,6 +123,15 @@ def load_fits(self, file: str | os.PathLike) -> None:
             The FITS file to load. If a string, it can be a URL or a
             file path.
         """
+        if isinstance(file, (str, os.PathLike)):
```
pllim, line 126: Do we not need to check for file extension? I guess maybe not for a dummy viewer, but probably necessary for actual implementation.

src/astro_image_display_api/interface_definition.py
```diff
@@ -52,41 +52,19 @@ class ImageViewerInterface(Protocol):
...
         """
-        Load a FITS file into the viewer.
+        Load data into the viewer. At a minimum, this should allow a FITS file
```
pllim, line 59: Depending on how the Jdaviz discussions go, we could also make this more flexible, like "if you already load data into the app another way, this method would add that data to the viewer this method is attached to" or something like that.

src/astro_image_display_api/dummy_viewer.py
mwcraig: Yeah, the dummy tries to do nothing substantive...

src/astro_image_display_api/interface_definition.py
pllim: I am thinking more like, people would still use `jdaviz.load(...)` outside of this API scope to read data to the app. Then then `jdaviz.viewers["blah"].load(data_label)` to put the data into the viewer with this API. So this means:

* Jdaviz does not need to worry about fitting this API at app level and limit its scope to viewer level.
* This API in Jdaviz only takes a string and would emit exception to tell user to use app level load otherwise.

Of course the exception part is not very elegant. You can also have `viewer.load` try to be smart and call app load, as you said above, but it increases the complexity of the method.

pllim: A more controversial suggestion is that we recognize that different backends might have very different ways to load data, and actually drop `load` from the API itself. This means the API here would just assume the data is already available. 🔥

src/astro_image_display_api/dummy_viewer.py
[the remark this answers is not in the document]
mwcraig: Sure, just added it

src/astro_image_display_api/dummy_viewer.py
```diff
@@ -99,6 +99,15 @@ def load_fits(self, file: str | os.PathLike) -> None:
...
+            self._load_fits(file)
```
pllim, line 103: Try to use `_load_asdf` here?
```suggestion
        if isinstance(file, (str, os.PathLike)):
            if isinstance(file, str):
                is_adsf = file.endswith(".asdf")
            else:
                is_asdf = file.suffix == ".asdf"
            if is_asdf:
                self._load_asdf(file)
            else:
                self._load_fits(file)
```

pllim (approved):
LGTM. Thanks!

I'll leave the merge to you.

PR #45, opened 2025-05-28
Implement catalog part of AIDA

### LINKED ISSUE #6: What happens if `remove_markers` is called with `marker_name=None`?
The API documentation (which is currently just the docstrings) does not say what happens when `remove_markers` is called with no `marker_name`. Could just be a noop.

### LINKED ISSUE #28: Change marker from a dictionary to a class or `namedtuple`, and change the name from `marker` to `marker_style`
Right now I have to either read the source code or set `iw.marker` to a value I know is bad so that I can be reminded of what the keys/values are for a marker.

If we changed it to a class (or a named tuple) then I could also introspect to get the attributes.

I would also find it clearer if the name were `marker_style`. 

Finally, in an ideal world, we might agree on a few basic shapes and colors that viewers implementing the interface need to support.

### LINKED ISSUE #29: Consider making `marker` private and instead only set it in `add_markers` or `start_marking`
Right now there's a mix of a stateful and a non-stateful idiom in how you set the style of markers.  `self.markers` is settable by the user, but it can also be set in `start_marking`.  `add_markers` also uses `self.markers`, but does *not* have a `markers` keyword.

So I see two ways to fix this:
1. Add a `markers` keyword to `add_markers`
2. Do 1 and *also* make `self.markers` private
3. remove `markers` from `start_marking`

I really don't like 3 because I think it just makes sense to be able to set your marker style when you add the marks.  So that leaves 1 and 2 - 2 is more "pure" (i.e., there's then one way to do it), but has the downside of breaking backwards compatibility (which we may or may not care about). 

Triggered by spacetelescope/jdaviz#699

cc @pllim @mwcraig

### LINKED ISSUE #32: Allow one to add markers from objects other than a table
While it is great that a table is an option, I often find myself creating a table just to use in add_markers. Accepting something like a Table or a numpy array would be convenient.

### LINKED ISSUE #40: Backend marker property comparisons
# Comparison of marker abilities in some backends

## Marker shapes

| Working name | matplotlib | bqplot | ginga | jdaviz |
| :--------: | :-------: | :------: | :---: | :----: |
| circle | "o" | "circle" | `Circle` object | ?? |
| square | "s" | "square" | `Square` object | ?? |
| crosshair (×) | "x" | "crosshair" | `Crosshair` object | ?? |
| cross or plus (+) | "+" | "plus" | ?? | ?? |
| diamond | "D" | "diamond" | ?? (maybe a rotated square?) | ?? |

### CSS color list: https://developer.mozilla.org/en-US/docs/Web/CSS/named-color

+ `matplotlib`
    + Hex RGB or RGBA
    + X11/CSS4 color names (case insensitive)
    + many others
+ `bqplot`
    + Gets its colors from the `Color` class in `ipywidgets`
    + Hex RGB or RGBA
    + This [bespoke list of colors hardcoded in ipywidgets](https://github.com/jupyter-widgets/ipywidgets/blame/57ff79dce80b38b6666f66d1c7436153fe7c85b5/python/ipywidgets/ipywidgets/widgets/trait_types.py#L13). This may be the CSS4 list, not sure though.
+ `ginga`
    + Hex RGB or RGBA
    + [List of color names](https://github.com/ejeschke/ginga/blob/main/ginga/colors.py#L15). This may be the CSS4 list or an extension of it (there are multiple shades of each color).
+ `jdaviz`
    + ??? but guessing hex and web colors

## Marker sizes

Does each backend have a way to set the marker size?

+ `matplotlib`
    + Yes, with the `markersize` parameter -- size is marker area or something like that
+ `bqplot`
    + Yes, with the `size` parameter -- size is marker area or something like that
+ `ginga`
    + Yes, marker size is radius or box side
+ `jdaviz`
    + Yes??

mwcraig:
This implements several things related to catalogs:

1. Change `add_markers`/`get_markers` to `load_catalog`/`get_catalog`.
2. Change `marker_name` to `catalog_label`
3. Add `set_catalog_style`/`get_catalog_style`
4. Made it an error to `set_catalog_style` before the catalog has been loaded. 
5. Made it an error to `get_catalog` or `get_catalog_style` without a `catalog_label` if there is more than one catalog.
6. Test fairly thoroughly for the ability to set/get without using catalog labels if there is only one catalog.
7. Made it an error to `get_catalog` if the provided `catalog_label` is a list.
8. Add and test `get_catalog_names`.

Edit:

- Fixes #28 
- Fixes #29
- Fixes #32 
- Fixes #40 
- Fixes #6

### DIFF
--- a/src/astro_image_display_api/dummy_viewer.py
+++ b/src/astro_image_display_api/dummy_viewer.py
@@ -365,7 +365,7 @@ class ImageViewer:
         try:
             del self._catalogs[catalog_label]
         except KeyError:
-            raise ValueError(f"Marker name {catalog_label} not found.")
+            raise ValueError(f"Catalog label {catalog_label} not found.")
 
     def get_catalog(self, x_colname: str = 'x', y_colname: str = 'y',
                     skycoord_colname: str = 'coord',
--- a/src/astro_image_display_api/dummy_viewer.py
+++ b/src/astro_image_display_api/dummy_viewer.py
@@ -171,20 +171,17 @@ class ImageViewer:
         **kwargs
             Additional keyword arguments to pass to the marker style.
         """
-        shape = shape
-        color = color
-        size = size
-
         catalog_label = self._resolve_catalog_label(catalog_label)
 
         if self._catalogs[catalog_label].data is None:
             raise ValueError("Must load a catalog before setting a catalog style.")
 
-        self._catalogs[catalog_label].style = {
-            "shape": shape,
-            "color": color,
-            "size": size,
-        }
+        self._catalogs[catalog_label].style = dict(
+            shape=shape,
+            color=color,
+            size=size,
+            **kwargs
+        )
 
     # Methods for loading data
     def load_image(self, file: str | os.PathLike | ArrayLike | NDData) -> None:
--- a/src/astro_image_display_api/dummy_viewer.py
+++ b/src/astro_image_display_api/dummy_viewer.py
@@ -1,4 +1,6 @@
 import os
+from collections import defaultdict
+from copy import deepcopy
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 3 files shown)


mwcraig:
> @ejeschke too?

Yes, please!

mwcraig:
FYI, rebasing to pull in changes from #44...

PR #46, opened 2025-05-30
Implement set_viewport/get_viewport in interface, tests, and dummy viewer

### LINKED ISSUE #8: Proposal: Add way to get center of viewer
There are a couple of ways to set/modify the position in the image of the center of the viewer (`center_on`, `offset_by`) but no way to retreive the position of the center. It seems like there should be. away to do that.

### LINKED ISSUE #10: What does zoom level mean when viewer has no concept of pixels?
Currently zoom level is defined in pixel space. However there might be situations where pixel space is meaningless in a viewer. Examples (non-exhaustive):

1. Imviz when linked by WCS uses a fake 10x10 hidden image as reference WCS. This causes zoom level value calculated in pixel space to be useless.
2. Sky viewer like WWT or Aladdin that has no concept of pixel space, only the sky.

What should we do:

* Allow these viewers to return `None`?
* Allow zoom level to be calculated in sky (arcsec, arcmin, arcdeg)? How?

mwcraig:
+ Fixes #10
+ Fixes #8

### DIFF
--- a/src/astro_image_display_api/dummy_viewer.py
+++ b/src/astro_image_display_api/dummy_viewer.py
@@ -308,7 +308,7 @@ class ImageViewer:
         """
         return int(shape[1] > shape[0])
 
-    def _intialize_image_viewport_stretch_cuts(
+    def _initialize_image_viewport_stretch_cuts(
         self,
         image_data: ArrayLike | NDData | CCDData,
         image_label: str | None,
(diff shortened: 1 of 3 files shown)


mwcraig:
Need to add image_label to set/get stretch and cuts...

mwcraig:
One more yolo...

PR #56, opened 2025-06-26
Change definition of FOV

### LINKED ISSUE #55: Consider changing definition of fov
`fov` in `get_viewport` and `set_viewport` is currently:

```
The field of view (FOV) of the viewport. If not given, the current FOV is used. If a float is given, it is interpreted a size in pixels. For images that are not square, the FOV is interpreted as the size of the longer side of the image.
```

I propose two changes:

1. Change the terminology for "For images that are not square... of the image" to "For viewers that are not square... of the viewer", to avoid confusion that "images" may be referring to the entire image, not the applied axes limits.

2. "as the size of the longer side of the image [viewer]" to "as the size of the shorter side of the image [viewer] such that the FOV is guaranteed to be entirely visible regardless of the aspect ratio of the viewer".  In other words, viewers should "pad" axes limits on the longer axis rather than "crop" on the shorter.  See https://github.com/spacetelescope/jdaviz/pull/3622#discussion_r2157225898 for more context.

cc @bmorris3

kecnry:
Closes #55

### DIFF
--- a/src/astro_image_display_api/interface_definition.py
+++ b/src/astro_image_display_api/interface_definition.py
@@ -441,9 +441,10 @@ class ImageViewerInterface(Protocol):
             The center of the viewport. If not given, the current center is used.
         fov : `astropy.units.Quantity` or float, optional
             The field of view (FOV) of the viewport. If not given, the current FOV
-            is used. If a float is given, it is interpreted a size in pixels. For images
-            that are not square, the FOV is interpreted as the size of the longer side
-            of the image.
+            is used. If a float is given, it is interpreted a size in pixels. For viewers
+            that are not square, the FOV is interpreted as the size of the shorter side
+            of the viewer such that the FOV is guaranteed to be entirely visible
+            regardless of the aspect ratio of the viewer.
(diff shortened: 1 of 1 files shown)


src/astro_image_display_api/interface_definition.py
```diff
@@ -441,9 +441,10 @@ def set_viewport(
             The center of the viewport. If not given, the current center is used.
         fov : `astropy.units.Quantity` or float, optional
             The field of view (FOV) of the viewport. If not given, the current FOV
-            is used. If a float is given, it is interpreted a size in pixels. For images
-            that are not square, the FOV is interpreted as the size of the longer side
-            of the image.
+            is used. If a float is given, it is interpreted a size in pixels. For viewers
```
bmorris3, line 444:
```suggestion
            is used. If a float is given, it is interpreted as a size in pixels. For viewers
```

bmorris3 (approved):
I think this is a good idea.

mwcraig (approved):
Thanks, this is better!

PR #75, opened 2025-07-08
Add kwargs to every method in AIDA and test that implementations allow kwargs

mwcraig:


src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -129,15 +129,24 @@ def _default_catalog_style(self) -> dict[str, Any]:
             "size": 5,
         }
 
-    def get_stretch(self, image_label: str | None = None) -> BaseStretch:
+    def get_stretch(
```
pllim, line 132: Do `get` stuff really need `**kwargs`? I thought we want flexibility to set things but get should be straightforward? Am I missing something?

Same comment on all the other `get` methods.

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -685,7 +733,10 @@ def set_viewport(
     set_viewport.__doc__ = ImageViewerInterface.set_viewport.__doc__
 
     def get_viewport(
-        self, sky_or_pixel: str | None = None, image_label: str | None = None
+        self,
+        sky_or_pixel: str | None = None,
+        image_label: str | None = None,
+        **kwargs,  # noqa: ARG002
```
pllim, line 739: That said, maybe we do want `**kwargs` here since viewer in Jdaviz has reference name as such, that might not apply to other backends.

pllim (commented):
Thanks!

I have comment on `get` but not sure if I left them at the correct place. Hope it is not too confusing.

src/astro_image_display_api/image_viewer_logic.py
mwcraig: I wondered the same thing....definitely `get_catalog_names` and `get_image_labels` will get turned into properties without an arguments once this is merged.

PR #86, opened 2026-06-22
Add test of viewport initialization and fix error

mwcraig:
The initial viewport size was being set incorrectly in the sample implementation included in this repository. This pull request adds a test for the issue to api_test and fixes the issue.


mwcraig:
This doesn't change the API, and ought to be uncontroversial, so merging...

PR #87, opened 2026-06-22
Add test to enforce get_viewport return value promise

mwcraig:
The documentation for get_viewport say that if `sky_or_pixel` is `None` then the return value should be in world coordinates if the image has a WCS. That was not the current behavior in the image viewer logic in this repo. A test for the return value has been added and the fix has been made.


mwcraig:
Another make-the-code-do-what-the-api-says fix, mergning.

Issue #106, opened 2026-07-07
Tracking: findings from the 2026-07 code review


mwcraig:
This issue tracks the findings of a critical review of the package (source, tests, docs) done in July 2026. Bugs that can silently affect science results are listed first. Each linked issue contains a reproduction script or permalinks to the relevant code at 7f8e611.

**High priority — science-affecting bugs**
- [x] #89 FOV ↔ pixel-scale conversions index the wrong axis of `proj_plane_pixel_scales` (H1)
- [x] #90 `get_catalog` broken: crash on fresh viewer, KeyError/silent no-op with column names, returns internal table by reference (H2)
- [x] #91 Catalog pixel↔sky conversions use the last-loaded image's WCS (H3)
- [x] #92 Unlabeled images/catalogs silently lost or shadowed; "unique name" not implemented (H4)

**Medium priority — correctness, test-suite validity, packaging**
- [x] #93 `get_viewport` sky/pixel default keyed to last-loaded image's WCS (M1)
- [x] #94 Operations on an empty viewer silently succeed (M2)
- [x] #95 `get_catalog_style`/`set_catalog_style` pollute the registry via `defaultdict` (M3)
- [x] #96 `remove_catalog("*")` skips default-state setup (M4)
- [x] #97 Initial viewport center off by half a pixel (M5)
- [x] #98 api_test contradicts the interface contract (M6)
- [x] #99 api_test weak assertions and coverage gaps (M7)
- [x] #100 Packaging: `__version__` always empty; license metadata mismatch (M8)
- [x] #101 Deduplicate label-resolution helpers, fix error wording (M9)

**Low priority — docs and polish**
- [x] #102 `get_cuts` return annotation contradicts interface (L1)
- [x] #103 Docstring drift (`set_catalog_style`, stale `remove_catalog` doc) (L2)
- [x] #104 Testing docs reference nonexistent `ImageDisplayInterface` (L3)
- [x] #105 Duplicate docstring test in `api_test.py` and `tests/` (L4)

Suggested ordering note: M6 (fixing api_test's empty-viewer tests) should land before or with M2, since two api_test tests currently depend on the buggy behavior.

---
*From a code review of `main` @ [`7f8e611`](https://github.com/astropy/astro-image-display-api/commit/7f8e611994deda64244ca49f990b5f8a1cb0ed69); reproductions were run against that commit.*

mwcraig:
**Second pass (2026-07): API-contract findings**

[...]

PR #117, opened 2026-07-17
Repair the image/catalog label model

### LINKED ISSUE #92: Unlabeled images/catalogs are silently lost or shadowed; the "unique name will be generated" promise is not implemented
Data loaded without a label is stored under the key `None`, which (a) never appears in `image_labels`/`catalog_labels`, (b) is silently **overwritten** by the next unlabeled load, and (c) becomes permanently unreachable once any labeled item exists, because `_resolve_image_label(None)` then resolves to the user label.

**Reproduction:**
```python
v = ImageViewerLogic()
v.load_image(arr1)                       # no label
print(v.image_labels)                    # () — invisible
v.load_image(arr2, image_label="second")
v.get_image() is arr2                    # True; arr1 is now unreachable by any call

v2 = ImageViewerLogic()
v2.load_catalog(c1); v2.load_catalog(c2)
print(v2.catalog_labels)                 # () — c1 silently overwritten by c2
```

The interface docstring for `load_catalog` says "If not given, a unique name will be generated" ([interface_definition.py#L368-L370](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L368-L370)) — the reference implementation does not do this, and the same policy question exists for `load_image` (where the interface is silent). Relevant code: [image_viewer_logic.py#L269-L304](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L269-L304).

Suggested direction: decide the semantics at the interface level (generate a real label, or make the default label a first-class visible label), then make labels round-trip: anything loadable must appear in `image_labels`/`catalog_labels` and stay reachable.

---

### LINKED ISSUE #94: Operations on an empty viewer silently succeed; the "image not found" checks are dead for the default label
`_set_up_catalog_image_dicts` seeds `_images`/`_catalogs` with a `None` entry ([image_viewer_logic.py#L84-L94](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L84-L94)), so every `if image_label not in self._images: raise ValueError(...)` guard passes when nothing is loaded:

```python
v = ImageViewerLogic()
v.get_stretch()            # None (interface documents ValueError for unloaded images)
v.get_cuts()               # None (interface return type: BaseInterval)
v.set_colormap("viridis")  # succeeds with no image
```

The interface documents `ValueError` when "the image_label does not correspond to a loaded image" (e.g. [interface_definition.py#L158-L162](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L158-L162)). Note fixing this requires fixing M6 first — two api_test tests currently *depend* on the dead-check behavior.

---

### LINKED ISSUE #95: `get_catalog_style`/`set_catalog_style` with an unknown label pollute the catalog registry
Because `_catalogs` is a `defaultdict`, merely *reading* a style creates a phantom catalog ([image_viewer_logic.py#L229-L238](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L229-L238)); the docstring instead promises an empty dict for unknown labels ([interface_definition.py#L443-L448](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L443-L448)).

```python
v = ImageViewerLogic()
v.load_catalog(c1, catalog_label="real")
v.get_catalog_style(catalog_label="phantom")  # {'catalog_label': 'phantom'} (not empty dict)
print(v.catalog_labels)                       # ('real', 'phantom')  <- phantom created
v.get_catalog_style()
# ValueError: Multiple catalog styles defined. ...  <- read call broke later calls
```
`set_catalog_style` with an unknown label raises correctly but still leaves the phantom entry behind ([image_viewer_logic.py#L248-L251](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L248-L251)).

---

### LINKED ISSUE #96: `remove_catalog("*")` leaves the viewer in a different state than a fresh viewer
[image_viewer_logic.py#L552-L556](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L552-L556) resets `_catalogs` to a bare `defaultdict`, skipping the default-entry/style setup done at init:

```python
v.remove_catalog(catalog_label="*")
v.get_catalog_style()   # {'catalog_label': None} — shape/color/size gone
# fresh viewer returns {'shape': 'circle', 'color': 'red', 'size': 5, 'catalog_label': None}
```

---

### LINKED ISSUE #101: Redundancy: `_resolve_image_label` and `_resolve_catalog_label` are copy-pasted near-duplicates (with copy-paste error messages)
[image_viewer_logic.py#L102-L125](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L102-L125) and [#L269-L292](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L269-L292) differ only in which dict they consult and their (both wrong) error strings: the catalog version always says "Multiple catalog **styles** defined … to get the **style**" even when resolving labels for data operations, and the image version says "… to get the style" too — a copy-paste remnant. One shared helper parameterized on the dict + a noun would remove the duplication and fix the messages. This same helper is where H4's default-label semantics should be implemented once.

---

### LINKED ISSUE #113: API design: missing-label behavior is opposite for images (raise) vs catalogs (return empty result)
This one is a design decision rather than a bug: the two halves of the API handle a label that doesn't correspond to loaded data in opposite ways, and the catalog half isn't even internally consistent. Filing it so maintainers can pick a policy deliberately; both options below are defensible.

**Current state (all contract-level, i.e. docstring and/or test-enforced):**

| Operation | Missing label → |
| --- | --- |
| `get_image`, `get_viewport`, `get_stretch`/`get_cuts`/`get_colormap`, `set_*` (image half) | **raises** `ValueError` "Image label ... not found" ([api_test.py#L181-L189](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L181-L189), [#L787-L797](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L787-L797)) |
| `get_catalog` | **returns an empty table** ([interface_definition.py#L529-L533](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L529-L533); test [api_test.py#L585-L588](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L585-L588)) |
| `get_catalog_style` | documented to **return an empty dict** ([interface_definition.py#L443-L448](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L443-L448)) |
| `remove_catalog` | **raises** `ValueError` ([interface_definition.py#L487-L492](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L487-L492); test [api_test.py#L676-L678](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L676-L678)) |

There is also an asymmetry *within* the style pair: `set_catalog_style` before any catalog is loaded must raise ([api_test.py#L419-L425](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L419-L425)) while `get_catalog_style` in the same state must succeed with defaults ([api_test.py#L427-L434](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L427-L434)). And per `load_catalog`'s Raises section, loading a *new* unlabeled catalog once two catalogs exist is an error ([interface_definition.py#L379-L384](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L379-L384)) — a load-time consequence of the same label-resolution policy.

**Option A — empty results everywhere (catalog-style semantics):** friendly to plotting/overlay loops (`for label in ...: plot(get_catalog(label))` never blows up), idempotent, no try/except in exploratory sessions. Cost: typos in labels fail silently, and "empty result" is indistinguishable from "loaded an empty catalog".

**Option B — raise everywhere (image-style semantics):** typos and stale labels surface immediately at the call site; `ValueError "... not found"` is already the established pattern on the image half, so it's the smaller change to the spec. Cost: callers that legitimately probe for optional data must catch exceptions or check `catalog_labels` first.

Either choice (or an explicit, documented rationale for why images and catalogs differ) would resolve this; the main ask is that the interface pick one policy and the docstrings + `api_test.py` enforce it uniformly, including the get/set style asymmetry.

Related: #94 (the not-found checks are dead for the default label — an implementation bug) and #92 (unlabeled data visibility). Neither frames the raise-vs-empty question as an interface-level decision.

---
*From a code review of `main` @ [`7f8e611`](https://github.com/astropy/astro-image-display-api/commit/7f8e611994deda64244ca49f990b5f8a1cb0ed69); this second pass focused on the API contract (interface docstrings vs `api_test.py` vs design).*

mwcraig:
> [!NOTE]
> This PR is stacked on #116 (`state-fixes-quick`); the diff shown here includes the commits from #116 until that PR merges. Only the "Repair the image/catalog label model" commit is new.

This PR repairs the label model in `ImageViewerLogic`, addressing the label-related findings from the July review (#106):

* **#101 — one label resolver.** `_resolve_image_label` and `_resolve_catalog_label` are now thin wrappers around a single parameterized `_resolve_label(label, kind, allow_new)` helper. The copy-pasted error messages ("Multiple catalog **styles** defined ... to get the **style**") are fixed as a side effect; the resolver also now owns the "label not found" check, replacing the previously dead per-method guards.

* **#92 — unlabeled loads get generated unique labels.** The interface already promised "a unique name will be generated" for `load_catalog`; this implements that promise for both catalogs and images. Data loaded without a label gets a label like `image-1` / `catalog-1`, appears in `image_labels`/`catalog_labels`, and can never be silently shadowed or lost. (`load_image`'s docstring now documents the same behavior, and `load_catalog` no longer raises for unlabeled loads with multiple catalogs present.)

* **#113 — unknown explicit labels raise everywhere.** Per the raise-for-both-halves policy, `get_catalog` and `get_catalog_style` now raise `ValueError` for a label that does not correspond to a loaded catalog, matching the image half, instead of returning an empty table / default dict. With *no* label and *nothing* loaded, `get_catalog` still returns an empty table and `get_catalog_style` still returns the default style (both are documented and test-enforced), so "is there anything here?" probes stay cheap.

* **#95 — reads no longer pollute state.** The `defaultdict` registries are plain dicts now; entries are created only by `load_image`/`load_catalog`, so a failed lookup can no longer create a phantom catalog that breaks later label-free calls.

* **#96 — `remove_catalog("*")` restores fresh-viewer state.** Removing all catalogs resets the registry to exactly the state a new viewer starts in, including the default catalog style.

* **#94 — empty-viewer guards are real.** The seeded `None` entries are gone, so `get_stretch`/`get_cuts`/`set_colormap`/`get_viewport`/etc. on a viewer with no image now raise `ValueError` as the interface documents, instead of silently succeeding.

Each fix comes with a regression test in `api_test.py` that fails on the previous implementation, and the interface docstrings have been updated where the contract changed.

Closes #92
Closes #94
Closes #95
Closes #96
Closes #101
Closes #113

### DIFF
--- a/src/astro_image_display_api/image_viewer_logic.py
+++ b/src/astro_image_display_api/image_viewer_logic.py
@@ -28,8 +28,10 @@ __all__ = ["ImageViewerLogic"]
 #: Label used for an image or catalog that is loaded without an explicit
 #: label. There is only ever one unlabeled image and one unlabeled catalog,
 #: so loading without a label repeatedly replaces the previous unlabeled
-#: image or catalog rather than accumulating new entries.
-DEFAULT_LABEL = "default"
+#: image or catalog rather than accumulating new entries. The value is
+#: deliberately unlikely to collide with a label a user would choose
+#: themselves.
+DEFAULT_LABEL = "_internal_default_label"
 
 
 @dataclass
@@ -102,10 +104,6 @@ class ImageViewerLogic:
         """
         Resolve a user-provided image or catalog label.
 
-        This is needed so that the user gets what they expect in the simple
-        case where there is only one image or catalog loaded. In that case
-        the user may or may not have actually specified a label.
-
         Parameters
         ----------
         label : str or None
@@ -130,6 +128,12 @@ class ImageViewerLogic:
             no label is given, if nothing is loaded or if several labels
             exist so the choice is ambiguous. Never raised when ``allow_new``
             is True.
+
+        Notes
+        -----
+        This is needed so that the user gets what they expect in the simple
+        case where there is only one image or catalog loaded. In that case
+        the user may or may not have actually specified a label.
         """
         registry, article = (
             (self._images, "an image")
@@ -587,7 +591,11 @@ class ImageViewerLogic:
             result = Table(names=["x", "y", "coord"])
         else:
             catalog_label = self._resolve_catalog_label(catalog_label)
-            result = self._catalogs[catalog_label].data
+            # Copy before renaming: rename_columns is in-place, and the
+            # table stored here is the actual registry entry, not a copy.
+            # Renaming it directly would permanently rename the columns of
+            # the stored catalog as a side effect of merely reading it.
+            result = self._catalogs[catalog_label].data.copy()
 
         result.rename_columns(
             ["x", "y", "coord"], [x_colname, y_colname, skycoord_colname]
--- a/src/astro_image_display_api/image_viewer_logic.py
+++ b/src/astro_image_display_api/image_viewer_logic.py
@@ -1,6 +1,5 @@
 import numbers
 import os
-from collections import defaultdict
 from copy import copy, deepcopy
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -82,47 +81,109 @@ class ImageViewerLogic:
         self._set_up_catalog_image_dicts()
 
     def _set_up_catalog_image_dicts(self):
-        # This is a dictionary of marker sets. The keys are the names of the
-        # marker sets, and the values are the tables containing the markers.
-        self._catalogs = defaultdict(CatalogInfo)
-        self._catalogs[None].data = None
-        self._catalogs[None].style = self._default_catalog_style.copy()
+        # Keys are the user-visible labels of the loaded catalogs and images.
+        # Entries are created only by the load_* methods, so reading state
+        # can never create an entry.
+        self._catalogs: dict[str, CatalogInfo] = {}
+        self._images: dict[str, ViewportInfo] = {}
 
-        self._images = defaultdict(ViewportInfo)
-        self._images[None].center = None
-        self._images[None].fov = None
-        self._images[None].wcs = None
+    @staticmethod
+    def _generate_label(registry: dict, kind: str) -> str:
+        """
+        Generate a unique label for a new image or catalog.
 
-    def _user_catalog_labels(self) -> list[str]:
-        """
-        Get the user-defined catalog labels.
-        """
-        return [label for label in self._catalogs if label is not None]
+        Parameters
+        ----------
+        registry : dict
+            The dictionary of already-loaded images or catalogs.
+        kind : str
+            Either ``"image"`` or ``"catalog"``; used as the prefix of the
+            generated label.
 
-    def _resolve_catalog_label(self, catalog_label: str | None) -> str:
+        Returns
+        -------
+        str
+            A label of the form ``"<kind>-<number>"`` that is not already
+            a key of ``registry``.
         """
-        Figure out the catalog label if the user did not specify one. This
-        is needed so that the user gets what they expect in the simple case
-        where there is only one catalog loaded. In that case the user may
-        or may not have actually specified a catalog label.
-        """
-        user_keys = self._user_catalog_labels()
-        if catalog_label is None:
-            match len(user_keys):
-                case 0:
-                    # No user-defined catalog labels, so return the default label.
-                    catalog_label = None
-                case 1:
-                    # The user must have loaded a catalog, so return that instead of
-                    # the default label, which live in the key None.
-                    catalog_label = user_keys[0]
-                case _:
-                    raise ValueError(
-                        "Multiple catalog styles defined. Please specify a "
-                        "catalog_label to get the style."
-                    )
+        index = len(registry) + 1
+        while (label := f"{kind}-{index}") in registry:
+            index += 1
+        return label
 
-        return catalog_label
+    def _resolve_label(
+        self,
+        label: str | None,
+        kind: str,
+        allow_new: bool = False,
+    ) -> str:
+        """
+        Resolve a user-provided image or catalog label.
+
+        This is needed so that the user gets what they expect in the simple
+        case where there is only one image or catalog loaded. In that case
+        the user may or may not have actually specified a label.
+
+        Parameters
+        ----------
+        label : str or None
+            The label the user provided, or None if they did not provide one.
+        kind : str
+            Either ``"image"`` or ``"catalog"``; selects which registry the
+            label is resolved against and is used in error messages.
+        allow_new : bool, optional
+            If True the label is being resolved for a load operation, so an
+            explicit label need not already exist and a missing label leads
+            to a unique generated label instead of an error.
+
+        Returns
+        -------
+        str
+            The resolved label.
+
+        Raises
+        ------
+        ValueError
+            If an explicit label does not correspond to loaded data, or, when
+            no label is given, if nothing is loaded or if several labels
+            exist so the choice is ambiguous. Never raised when ``allow_new``
+            is True.
+        """
+        registry, article = (
+            (self._images, "an image")
+            if kind == "image"
+            else (self._catalogs, "a catalog")
+        )
+
+        if label is not None:
+            if not allow_new and label not in registry:
+                raise ValueError(
+                    f"{kind.capitalize()} label '{label}' not found. "
+                    f"Please load {article} first."
+                )
+            return label
+
+        if allow_new:
+            return self._generate_label(registry, kind)
+
+        match len(registry):
+            case 0:
+                raise ValueError(f"No {kind} is loaded. Please load {article} first.")
+            case 1:
+                return next(iter(registry))
+            case _:
+                raise ValueError(
+                    f"Multiple {kind} labels defined. Please specify a "
+                    f"{kind}_label to select one."
+                )
+
+    def _resolve_catalog_label(
+        self, catalog_label: str | None, allow_new: bool = False
+    ) -> str:
+        """
+        Resolve a user-provided catalog label; see `_resolve_label`.
+        """
+        return self._resolve_label(catalog_label, "catalog", allow_new=allow_new)
 
     @property
     def _default_catalog_style(self) -> dict[str, Any]:
@@ -141,10 +202,6 @@ class ImageViewerLogic:
         **kwargs,  # noqa: ARG002
     ) -> BaseStretch:
         image_label = self._resolve_image_label(image_label)
-        if image_label not in self._images:
-            raise ValueError(
-                f"Image label '{image_label}' not found. Please load an image first."
-            )
         return self._images[image_label].stretch
 
     def set_stretch(
@@ -159,10 +216,6 @@ class ImageViewerLogic:
                 "`astropy.visualization` Stretch object."
             )
         image_label = self._resolve_image_label(image_label)
-        if image_label not in self._images:
-            raise ValueError(
-                f"Image label '{image_label}' not found. Please load an image first."
-            )
         self._images[image_label].stretch = value
 
     def get_cuts(
@@ -171,10 +224,6 @@ class ImageViewerLogic:
         **kwargs,  # noqa: ARG002
     ) -> tuple:
         image_label = self._resolve_image_label(image_label)
-        if image_label not in self._images:
-            raise ValueError(
-                f"Image label '{image_label}' not found. Please load an image first."
-            )
         return self._images[image_label].cuts
 
     def set_cuts(
@@ -193,10 +242,6 @@ class ImageViewerLogic:
                 "of two values."
             )
         image_label = self._resolve_image_label(image_label)
-        if image_label not in self._images:
-            raise ValueError(
-                f"Image label '{image_label}' not found. Please load an image first."
-            )
         self._images[image_label].cuts = cuts
 
     def set_colormap(
@@ -206,10 +251,6 @@ class ImageViewerLogic:
         **kwargs,  # noqa: ARG002
     ) -> None:
         image_label = self._resolve_image_label(image_label)
-        if image_label not in self._images:
-            raise ValueError(
-                f"Image label '{image_label}' not found. Please load an image first."
-            )
         self._images[image_label].colormap = map_name
 
     def get_colormap(
@@ -218,10 +259,6 @@ class ImageViewerLogic:
         **kwargs,  # noqa: ARG002
     ) -> str:
         image_label = self._resolve_image_label(image_label)
-        if image_label not in self._images:
-            raise ValueError(
-                f"Image label '{image_label}' not found. Please load an image first."
-            )
         return self._images[image_label].colormap
 
     # The methods, grouped loosely by purpose
@@ -231,6 +268,13 @@ class ImageViewerLogic:
         catalog_label=None,
         **kwargs,  # noqa: ARG002
     ) -> dict[str, Any]:
+        if catalog_label is None and not self._catalogs:
+            # Nothing is loaded, so report the default style that a new
+            # catalog would get.
+            style = self._default_catalog_style
+            style["catalog_label"] = None
+            return style
+
         catalog_label = self._resolve_catalog_label(catalog_label)
 
         style = self._catalogs[catalog_label].style.copy()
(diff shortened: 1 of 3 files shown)


mwcraig:
This needs a major rewrite -- there should be only ONE unspecified label. Calling, e.g., `load_image` without an image label multiple times SHOULD results in replacing the image (same for catalog).

mwcraig:
[Claude, at Matt's direction] Good catch — you're right that generating `image-1`, `image-2`, ... for unlabeled loads was the wrong model. Fixed in 2d80f11 on `label-model-fixes`: `_generate_label` is gone, replaced by a single shared `DEFAULT_LABEL = "default"` that `_resolve_label` returns for any load without an explicit label. So `load_image()`/`load_catalog()` with no label now repeatedly replace the previous unlabeled image/catalog instead of piling up new entries; a labeled load still coexists with the default one. Explicit-label lookups, the resolve-to-sole-entry convenience, and the ValueError behavior for unknown/ambiguous labels are unchanged. Interface/logic docstrings updated and the two regression tests rewritten to pin replace-on-unlabeled-load. Propagated up the stack by merge (no rebase): catalog-wcs-fixes, template-hooks, docs-test-cleanup — full suites pass at every level (68/77/84/84), and astrowidgets' bqplot+ginga ImageAPITest subclasses pass against the updated tree with no backend changes needed.

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -26,6 +25,12 @@
 
 __all__ = ["ImageViewerLogic"]
 
+#: Label used for an image or catalog that is loaded without an explicit
+#: label. There is only ever one unlabeled image and one unlabeled catalog,
+#: so loading without a label repeatedly replaces the previous unlabeled
+#: image or catalog rather than accumulating new entries.
+DEFAULT_LABEL = "default"
```
mwcraig: Instead of this simple default, use something the user is highly unlikely to ever use themselves, e.g. "_internal_default_label"

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -82,47 +87,89 @@ def __post_init__(self):
...
+        self,
+        label: str | None,
+        kind: str,
+        allow_new: bool = False,
+    ) -> str:
         """
-        Get the user-defined catalog labels.
+        Resolve a user-provided image or catalog label.
+
+        This is needed so that the user gets what they expect in the simple
```
mwcraig: This should be in a Notes section

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Also, why is a string label better than `None` for this case?

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -82,47 +87,89 @@ def __post_init__(self):
...
+            # shared default label, so repeated unlabeled loads replace the
+            # previous unlabeled image or catalog rather than piling up new
+            # ones.
+            return DEFAULT_LABEL
+
+        match len(registry):
+            case 0:
+                raise ValueError(f"No {kind} is loaded. Please load {article} first.")
+            case 1:
+                return next(iter(registry))
```
mwcraig: Why? Isn't registry[0] easier?

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -82,47 +87,89 @@ def __post_init__(self):
...
+        # Entries are created only by the load_* methods, so reading state
+        # can never create an entry.
+        self._catalogs: dict[str, CatalogInfo] = {}
+        self._images: dict[str, ViewportInfo] = {}
+
+    def _resolve_label(
+        self,
+        label: str | None,
+        kind: str,
+        allow_new: bool = False,
```
mwcraig, line 102: Why is allow_new needed?

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -556,17 +559,16 @@ def remove_catalog(
             )
         elif catalog_label == "*":
             # If the user wants to remove all catalogs, we reset the
-            # catalogs dictionary to an empty one.
-            self._catalogs = defaultdict(CatalogInfo)
```
mwcraig, line 560: Why not defaultdict?

src/astro_image_display_api/interface_definition.py
```diff
@@ -379,9 +382,7 @@ def load_catalog(
         Raises
         ------
         ValueError
-            If the ``table`` does not contain the required columns, or if
-            the ``catalog_label`` is not provided when there are multiple
```
mwcraig, line 383: Why is the catalog_label part dropped here?

src/astro_image_display_api/interface_definition.py
```diff
@@ -366,8 +368,9 @@ def load_catalog(
             If `True`, the ``skycoord_colname`` column will be used to
             get the marker positions. Default is `False`.
         catalog_label : str, optional
-            The name of the marker set to use. If not given, a unique
-            name will be generated.
+            The name of the marker set to use. If not given, a single shared
```
mwcraig:
```suggestion
            The name to use for the catalog. If not given, a single shared
```

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -571,15 +592,16 @@ def get_catalog(
...
-            else Table(names=["x", "y", "coord"])
-        )
+        if catalog_label is None and not self._catalogs:
+            # Nothing is loaded; return an empty table with the expected
+            # columns rather than raising, so that "is there anything
+            # here?" queries are easy to write.
+            result = Table(names=["x", "y", "coord"])
+        else:
+            catalog_label = self._resolve_catalog_label(catalog_label)
+            result = self._catalogs[catalog_label].data
```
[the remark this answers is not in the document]
mwcraig: [Claude, at Matt's direction] Fixed in 23d6d7f: `get_catalog` now does `self._catalogs[catalog_label].data.copy()` before `rename_columns`, so reads can no longer rename columns in the stored registry entry.

src/astro_image_display_api/interface_definition.py
```diff
@@ -460,7 +461,8 @@ def get_catalog_style(self, catalog_label: str | None = None, **kwargs) -> dict:
 
         ValueError
             If there are multiple catalog styles set and the user has not
-            specified a ``catalog_label`` for which to get the style.
+            specified a ``catalog_label`` for which to get the style, or if
+            the ``catalog_label`` does not correspond to a loaded catalog.
```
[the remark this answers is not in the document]
mwcraig: [Claude, at Matt's direction] Fixed in 23d6d7f: reworded to "multiple catalog labels defined" to match the actual `ValueError` message, since styles are keyed by catalog label rather than tracked independently.

src/astro_image_display_api/image_viewer_logic.py
mwcraig: [Claude, at Matt's direction] Fixed in 23d6d7f: `DEFAULT_LABEL` is now `"_internal_default_label"` instead of `"default"`. Also updated the one test (`test_unlabeled_image_loads_share_one_default_label`) that hardcoded the old literal `"default"` string to derive the label from `image_labels` instead, so it is not coupled to the constant's value.

src/astro_image_display_api/image_viewer_logic.py
mwcraig: [Claude, at Matt's direction] Fixed in 23d6d7f: moved the rationale paragraph out of the docstring summary and into a Notes section, per numpydoc convention.

src/astro_image_display_api/interface_definition.py
mwcraig: [Claude, at Matt's direction] Applied in 23d6d7f.

src/astro_image_display_api/image_viewer_logic.py
mwcraig: [Claude, at Matt's direction] A string is needed because the parameter itself already uses `None` to mean "caller did not pass a label" (`label: str | None = None`). If the registry also used `None` as a real key for "the unlabeled entry," those two meanings collide: you could no longer tell "user omitted the argument, resolve to whatever the default is" apart from "user is asking for the entry literally keyed by None." Using a distinct sentinel string keeps the two levels separate, and it also means the label shows up cleanly in `image_labels`/`catalog_labels` and in error messages without needing `None`-specific formatting everywhere it is displayed or iterated.

src/astro_image_display_api/image_viewer_logic.py
mwcraig: [Claude, at Matt's direction] `registry` is a `dict[str, ...]`, and dicts are not positionally indexable — `registry[0]` would look up the key `0`, not "the first item," and raise `KeyError` (unless a key literally equal to the int `0` existed). `next(iter(registry))` is the standard idiom for "give me the (only) key" without materializing a list via `list(registry)[0]`.

src/astro_image_display_api/image_viewer_logic.py
mwcraig: [Claude, at Matt's direction] `_resolve_label` is shared by two different call sites with different error semantics. Reads (`get_*`/`set_*`/`remove_*`) must raise on an unknown explicit label, on nothing loaded, and on an ambiguous omitted label when multiple entries exist. Loads (`load_image`/`load_catalog`) must never raise on the label itself: an explicit label is allowed to be new (that is how you create an entry), and an omitted label always resolves to the single shared default rather than erroring on "nothing loaded yet" or "multiple already loaded." `allow_new=True` is what selects the load-side behavior instead of the read-side behavior in the shared helper.

src/astro_image_display_api/image_viewer_logic.py
mwcraig: [Claude, at Matt's direction] This is the fix, not a stylistic change: the old code was `self._catalogs = defaultdict(CatalogInfo)` with `self._catalogs[None]` used as an "empty" sentinel, so simply reading `self._catalogs[some_label]` (e.g. in `get_catalog`/`get_catalog_style`) silently created a phantom entry as a side effect (this is the bug Copilot flagged elsewhere in this PR, and is also the root cause of #95). Switching to a plain `dict` and the comment above it ("Entries are created only by the load_* methods, so reading state can never create an entry") make that auto-vivification impossible by construction.

src/astro_image_display_api/interface_definition.py
mwcraig: [Claude, at Matt's direction] Dropped intentionally, not an oversight: with the current model, `load_catalog` resolves its label with `allow_new=True`, and per `_resolve_label`'s contract that path is "[n]ever raised" for label reasons — an explicit label is accepted even if new, and an omitted label always resolves to the shared default regardless of how many catalogs are already loaded. So `load_catalog` can no longer raise `ValueError` for an ambiguous/missing `catalog_label`; the only remaining `ValueError` there is for a `table` missing required columns. There's no test asserting the old ambiguous-label-on-load error, which matches this.

src/astro_image_display_api/image_viewer_logic.py
mwcraig: I don't ever anticipate the length of the registry being large -- go ahead and materialize that list.

mwcraig: [Claude, at Matt's direction] Done in 8911377: changed to `list(registry)[0]` for clarity.

Issue #114, opened 2026-07-15
Record the displayed image label in `load_image` and expose it (e.g. a `displayed_image_label` property)


mwcraig:
## Motivation

`ImageViewerLogic` stores per-label settings (cuts, stretch, colormap, viewport) for every loaded image, but a typical backend built on it displays one image at a time. That means every such backend needs the concept "which label is currently displayed" in order to decide whether a `set_*`/`get_*` call should touch the live view:

- `set_cuts(value, image_label=X)` should update the stored settings for `X` unconditionally, but push them into the live viewer only if `X` is the displayed image.
- Conversely, syncing live viewer state back into stored per-label state (e.g. reading the current pan/zoom in `get_viewport(image_label=X)`) is only valid when `X` is the displayed image.

This is not hypothetical: the astrowidgets ginga backend (astropy/astrowidgets#205) had confirmed cross-label clobbering bugs for exactly this reason. Calls like `set_cuts(..., image_label=X)` for a non-displayed `X` clobbered the displayed image's on-screen state, and `get_viewport(image_label=X)` for a non-displayed `X` corrupted `X`'s stored viewport by syncing it from the live pan/scale of a different image. The fix tracks a private `self._displayed_image_label`, set in `load_image`, and guards the apply/sync helpers on it.

Since every single-image-at-a-time backend needs the same guard, it would be better for `ImageViewerLogic` to provide the concept directly rather than each backend reinventing it privately.

## `ImageViewerLogic` already half-tracks this

`load_image` already records one piece of "the current image" implicitly: it sets `self._wcs` to the just-loaded image's WCS ([image_viewer_logic.py#L321-L323](https://github.com/astropy/astro-image-display-api/blob/main/src/astro_image_display_api/image_viewer_logic.py)). The concept exists in the implementation; it just isn't named or exposed.

## Proposal

Record the resolved label in `load_image` and expose it read-only:

```python
def load_image(self, file, image_label=None, **kwargs):
    image_label = self._resolve_image_label(image_label)
    ...
    self._displayed_image_label = image_label

@property
def displayed_image_label(self):
    """Label of the most recently loaded (displayed) image, or None if no
    image has been loaded."""
    return self._displayed_image_label
```

A backend can then guard its live-view updates:

```python
def set_cuts(self, value, image_label=None, **kwargs):
    super().set_cuts(value, image_label=image_label, **kwargs)
    if self._resolve_image_label(image_label) == self.displayed_image_label:
[...]

PR #118, opened 2026-07-17
Fix get_catalog defects; use per-image WCS for catalog conversions and get_viewport

### LINKED ISSUE #90: `get_catalog` is broken several ways: crash on fresh viewer, KeyError with custom column names, silent rename no-op, and returns the internal table by reference
Code: [image_viewer_logic.py#L566-L588](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L566-L588)

Four related defects:

1. **Crash on a fresh viewer.** The default `None` entry exists in `_catalogs` with `data=None`, so the `else Table(...)` branch is skipped and `None.rename_columns(...)` raises `AttributeError`. The interface promises an empty table ([interface_definition.py#L529-L533](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L529-L533)).
   ```python
   ImageViewerLogic().get_catalog()
   # AttributeError: 'NoneType' object has no attribute 'rename_columns'
   ```

2. **KeyError when the catalog was loaded with non-default column names.** `load_catalog` stores columns under the *user-provided* names, but `get_catalog` assumes the stored names are always `["x", "y", "coord"]`:
   ```python
   v = ImageViewerLogic()
   v.load_catalog(Table(dict(px=[1.0], py=[3.0])), x_colname="px", y_colname="py", catalog_label="c")
   v.get_catalog(x_colname="px", y_colname="py", catalog_label="c")
   # KeyError: 'Column x does not exist'
   ```

3. **Silent rename no-op → wrong column names returned.** Even in the default-name case, `rename_columns` silently does nothing, because the stored table was created with `deepcopy(table)` ([image_viewer_logic.py#L489](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L489)) and renaming a deep-copied `Table` updates the column's `info.name` without updating the table's column mapping (astropy 7.1.0 — arguably an astropy bug worth reporting upstream; `Table.copy()` does not have the problem):
   ```python
   v = ImageViewerLogic()
   v.load_catalog(Table(dict(x=[1.0], y=[2.0])), catalog_label="c")
   t = v.get_catalog(x_colname="xcen", y_colname="ycen", catalog_label="c")
   print(t.colnames)   # ['x', 'y', 'coord']  — silently NOT renamed
   ```

4. **Returned table is the stored table by reference** — user mutations corrupt the viewer's copy (and the in-place rename in (3), when it works, would too):
   ```python
   v = ImageViewerLogic()
   v.load_catalog(Table(dict(x=[1.0, 2.0], y=[3.0, 4.0])), catalog_label="c")
   t = v.get_catalog(catalog_label="c")
   t["x"][0] = -999.0
   print(v.get_catalog(catalog_label="c")["x"][0])   # -999.0 — internal state corrupted
   ```
   Note `api_test.test_load_multiple_catalogs` calls `.sort()` on the returned table, mutating viewer state mid-test ([api_test.py#L555-L557](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L555-L557)).

Suggested direction: store catalogs under canonical internal column names, return a **copy** renamed to the requested names, and make the empty-catalog branch actually reachable for the default label. Replace `deepcopy(table)` with `table.copy()`.

---

### LINKED ISSUE #91: Catalog pixel↔sky conversions use the WCS of the *last-loaded image* (global `self._wcs`)
`load_image` stores the most recent image's WCS in the shared attribute `self._wcs` ([image_viewer_logic.py#L321-L323](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L321-L323)), and `load_catalog` uses that WCS to compute x/y from sky coordinates and vice versa ([image_viewer_logic.py#L491-L512](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L491-L512)). With more than one image loaded, catalog pixel positions are silently computed with whichever image happened to load last; there is no way to associate a catalog with an image.

**Reproduction:**
```python
v = ImageViewerLogic()
v.load_image(NDData(data=np.zeros((100, 100)), wcs=w1), image_label="a")  # crpix=(50,50)
v.load_image(NDData(data=np.zeros((100, 100)), wcs=w2), image_label="b")  # crpix=(500,500)
v.load_catalog(Table(dict(coord=SkyCoord([150.0], [30.0], unit="deg"))), catalog_label="c")
print(v.get_catalog(catalog_label="c")["x"][0])   # 499.0 — computed with w2, even if the
                                                  # user meant the catalog for image "a"
```

This is also an API-design question: `load_catalog` has no `image_label` parameter, so the interface itself cannot express which image a catalog belongs to. Worth deciding at the interface level, not just in the reference implementation.

---

### LINKED ISSUE #93: `get_viewport` decides sky-vs-pixel default from the *last-loaded* image's WCS, not the requested image's
[image_viewer_logic.py#L690-L696](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L690-L696) tests `self._wcs` (set by the most recent `load_image`) instead of `viewport.wcs`. The docstring says the decision is per-image ([interface_definition.py#L628-L635](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L628-L635)).

**Reproduction:**
```python
v = ImageViewerLogic()
v.load_image(np.zeros((10, 10)), image_label="nowcs")
v.load_image(NDData(data=np.zeros((10, 10)), wcs=w), image_label="withwcs")
v.get_viewport(image_label="nowcs")
# ValueError: WCS is not set. Cannot convert pixel coordinates to sky coordinates.
```
In the reverse load order the failure is silent instead: the WCS image's viewport comes back in pixel coordinates when the contract says sky. Fix: use `viewport.wcs`.

---

mwcraig:
> [!NOTE]
> This PR is stacked on #116 and #117 — the diff includes their commits until they merge. Only the last commit (\`Fix get_catalog defects and use per-image WCS...\`) is new here.

This is the third of three PRs fixing state-model issues found in the July 2026 code review (#106 tracker).

### `get_catalog` fixes (#90)

- Catalog position columns are now stored under canonical internal names (`x`, `y`, `coord`), so a catalog loaded with custom column names is retrievable, and `get_catalog` renames a copy to whatever column names the caller requests. Previously the rename was a silent no-op (and a `KeyError` for custom-named catalogs).
- `load_catalog` copies the input with `Table.copy()` instead of `deepcopy()`; the deep copy left the table's column mapping out of sync after renames.
- `get_catalog` returns a **copy** of the stored table, so mutating the result can no longer corrupt viewer state, and the fresh-viewer empty table is built directly with the requested column names.

### Per-image WCS (#91)

The global `self._wcs`, which always pointed at the *last-loaded* image's WCS, is gone. `load_catalog` now picks the WCS for pixel↔sky conversion with the same defaulting rule already used for labels when none is given:

- exactly one image loaded → that image's WCS;
- no image loaded → no WCS (unchanged behavior);
- several images loaded → ambiguous: raises `ValueError` when a conversion is actually required (pixel columns must be computed from sky coordinates, or `use_skycoord=True` with no coordinate column), and otherwise skips the opportunistic enrichment of the missing coordinate column rather than filling it with an arbitrary image's WCS.

Once the displayed-image concept lands (#114), this rule can tighten to "the displayed image's WCS."

### `get_viewport` sky-vs-pixel default (#93)

`get_viewport` now decides its sky-vs-pixel default from the **requested** image's WCS (`viewport.wcs`) rather than global state, matching the documented per-image contract.

### Tests

Nine regression tests added to `api_test.py`, including two-image tests with different WCS verifying that catalog conversions refuse ambiguity instead of silently using the wrong WCS, that a loaded catalog's coordinates are not rewritten by later image loads, and that `get_viewport` picks the right image's WCS in both load orders. Full suite: 77 passed.

Closes #90
Closes #91
Closes #93

### DIFF
--- a/src/astro_image_display_api/api_test.py
+++ b/src/astro_image_display_api/api_test.py
@@ -778,6 +778,78 @@ class ImageAPITest:
         coords = wcs.pixel_to_world(catalog["x"], catalog["y"])
         assert all(coords.separation(retrieved_catalog["coord"]) < 1e-9 * u.deg)
 
+    def test_get_catalog_empty_on_fresh_viewer(self):
+        # Regression test for #90: get_catalog on a viewer with no catalog
+        # loaded must return an empty table with the requested column
+        # names rather than raising.
+        self._assert_empty_catalog_table(self.image.get_catalog())
+
+        tab = self.image.get_catalog(
+            x_colname="px", y_colname="py", skycoord_colname="sky"
+        )
+        assert len(tab) == 0
+        assert sorted(tab.colnames) == sorted(["px", "py", "sky"])
+
+    def test_load_get_catalog_custom_column_names(self, catalog):
+        # Regression test for #90: a catalog loaded with non-default column
+        # names must be retrievable with those names...
+        renamed = catalog.copy()
+        renamed.rename_columns(["x", "y", "coord"], ["px", "py", "sky"])
+        self.image.load_catalog(
+            renamed,
+            x_colname="px",
+            y_colname="py",
+            skycoord_colname="sky",
+            catalog_label="cat",
+        )
+
+        tab = self.image.get_catalog(
+            x_colname="px", y_colname="py", skycoord_colname="sky", catalog_label="cat"
+        )
+        assert sorted(tab.colnames) == sorted(["px", "py", "sky"])
+        np.testing.assert_allclose(tab["px"], catalog["x"])
+        np.testing.assert_allclose(tab["py"], catalog["y"])
+
+        # ...and get_catalog must return whatever column names are asked for.
+        tab = self.image.get_catalog(catalog_label="cat")
+        assert sorted(tab.colnames) == sorted(["x", "y", "coord"])
+        np.testing.assert_allclose(tab["x"], catalog["x"])
+
+    def test_get_catalog_renames_columns(self, catalog):
+        # Regression test for #90: renaming the position columns to the
+        # requested names must actually happen instead of being a silent
+        # no-op that returns the default names.
+        self.image.load_catalog(catalog, catalog_label="cat")
+
+        tab = self.image.get_catalog(
+            x_colname="xcen",
+            y_colname="ycen",
+            skycoord_colname="sky",
+            catalog_label="cat",
+        )
+        assert "xcen" in tab.colnames
+        assert "ycen" in tab.colnames
+        assert "sky" in tab.colnames
+        assert "x" not in tab.colnames
+        assert "y" not in tab.colnames
+        assert "coord" not in tab.colnames
+        np.testing.assert_allclose(tab["xcen"], catalog["x"])
+        np.testing.assert_allclose(tab["ycen"], catalog["y"])
+
+    def test_get_catalog_returns_copy(self, catalog):
+        # Regression test for #90: the returned table must be a copy, so
+        # that modifying it cannot corrupt the stored catalog.
+        self.image.load_catalog(catalog, catalog_label="cat")
+
+        tab = self.image.get_catalog(catalog_label="cat")
+        original_value = tab["x"][0]
+        tab["x"][0] = -999.0
+        tab.remove_column("y")
+
+        again = self.image.get_catalog(catalog_label="cat")
+        assert again["x"][0] == original_value
+        assert "y" in again.colnames
+
     def test_catalog_info_preserved_after_load(self, catalog):
         # Make sure that any catalog columns in addition to the position data
         # is preserved after loading a catalog.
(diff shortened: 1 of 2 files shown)


src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -474,6 +469,50 @@ def save(
         p.write_text("This is a dummy file. The viewer does not save anything.")
 
     # Marker-related methods
+    def _catalog_conversion_wcs(self, conversion_required: bool) -> WCS | None:
+        """
+        Return the WCS to use for catalog pixel↔sky coordinate conversion.
+
+        A catalog cannot yet be associated with a particular image, so the
```
mwcraig: Put this in Notes. Also, there is no plan to associate a catalog with a particular image.

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -495,10 +534,18 @@ def load_catalog(
         except KeyError:
             xy = None
 
-        to_add = deepcopy(table)
```
mwcraig, line 498: I know that deepcopy is being avoided here because it makes the `Table.rename_columns` fail. Can you give me a small reproducer of that issue?

src/astro_image_display_api/api_test.py
```diff
@@ -778,6 +778,78 @@ def test_load_catalog_with_xy_and_wcs(self, catalog, data, wcs):
...
+            y_colname="py",
+            skycoord_colname="sky",
+            catalog_label="cat",
+        )
+
+        tab = self.image.get_catalog(
+            x_colname="px", y_colname="py", skycoord_colname="sky", catalog_label="cat"
+        )
+        assert sorted(tab.colnames) == sorted(["px", "py", "sky"])
+        np.testing.assert_allclose(tab["px"], catalog["x"])
```
mwcraig: Shouldn't these be identical since no operations are done on the data?

src/astro_image_display_api/api_test.py
```diff
@@ -882,6 +954,86 @@ def test_adding_catalog_as_world(self, data, wcs):
             result["coord"].dec.deg, mark_coord_table["coord"].dec.deg
         )
 
+    @staticmethod
+    def _make_tan_wcs(crval, crpix, scale=0.001):
```
mwcraig, line 958: Is there already some like this elsewhere in the test file?

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Will do — the explanation will move into a numpydoc Notes section in the next push, and the "cannot yet be associated with a particular image" wording will be dropped (it implied a plan that does not exist); the Notes will just state the selection rule itself.

*This reply was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Sure — verified just now against astropy 7.0.2, and it does not even need a SkyCoord column:

```python
from copy import deepcopy
from astropy.table import Table

t = Table({"px": [1.0, 2.0], "py": [3.0, 4.0]})
td = deepcopy(t)
td.rename_columns(["px", "py"], ["x", "y"])

td.colnames                                  # ["px", "py"]  <- rename silently did not take
td["x"]                                      # KeyError: "x"
[c.info.name for c in td.columns.values()]   # ["x", "y"]    <- but the columns themselves WERE renamed
```

What happens: `rename_columns` works through each column's `info.name` setter, which notifies the parent table (via the column's `parent_table` weak reference) to re-key its `columns` mapping. After `deepcopy`, the copied columns' `parent_table` no longer points at the copied table — checking identity, it is neither the original nor the copy — so the notification goes nowhere: the columns' own `info.name` changes but the table's column mapping keeps the old keys. That is the "column mapping out of sync" failure: `colnames` still shows the old names and the new names raise `KeyError`. `Table.copy()` rebuilds the table properly, so the parent links are intact and the rename behaves. (The original table is unaffected either way.)

In the old code this was masked because the rename targets were usually already named `x`/`y`/`coord`, making the rename a no-op; it bit exactly when a caller used custom column names — the case #90 is about.

*This reply was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/api_test.py
mwcraig: Yes — nothing on this path does arithmetic on the pixel columns (the table is copied and the columns re-keyed; the values are untouched), so exact equality holds. `assert_allclose` was just copying the style of the neighboring catalog tests; will tighten this (and the matching `py` line) to `np.testing.assert_array_equal` in the next push.

*This reply was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/api_test.py
mwcraig: Similar but not quite. The file has the `wcs` fixture (the Airy-projection example from the astropy docs) and three inline TAN-WCS constructions in the FOV tests (`test_fov_anisotropic_pixel_scale`, `test_fov_sky_to_pixel_honors_quantity_unit`, `test_fov_wcs_with_arcsec_cunit`), but no parameterized helper. The new multi-image tests need *two different* WCSes with controllable `crval`/`crpix`, which the single fixture cannot supply — hence this helper.

The inline FOV constructions cannot use it as-is (they need anisotropic/per-axis pixel scales and, in one case, arcsec `cunit`), but `_make_tan_wcs` could grow a `cdelt` pair parameter and consolidate them. That refactor would slot naturally into the test-cleanup batch in #120 rather than this PR — happy to do it there.

*This reply was written by Claude at Matt Craig's direction.*

mwcraig: Add a note to 120 so we don't forget to do that.

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Done in 325f741.

src/astro_image_display_api/api_test.py
mwcraig: Done in 325f741 — also tightened the same pattern in `test_get_catalog_renames_columns` and the default-names retrieval check; `assert_allclose` remains only where values go through a WCS conversion.

src/astro_image_display_api/api_test.py
mwcraig: Noted on #120: https://github.com/astropy/astro-image-display-api/pull/120#issuecomment-5008436047

PR #119, opened 2026-07-17
Add template-method rendering hooks to ImageViewerLogic

### LINKED ISSUE #115: Provide a docs-from-interface decorator for backends
In astropy/astrowidgets#205, both the ginga and bqplot backends carry an identical copy of a small class decorator, `docs_from_super_if_missing`, that copies docstrings from the AIDA interface methods onto a backend's overriding methods when the override has no docstring of its own:

```python
def docs_from_super_if_missing(cls):
    for name, method in cls.__dict__.items():
        if not name.startswith("_"):
            if method.__doc__:
                continue
            interface_method = getattr(ImageViewerLogic, name, None)
            if interface_method:
                method.__doc__ = interface_method.__doc__
    return cls
```

Every AIDA backend that overrides interface methods (typically to push state into its viewer after calling `super()`) will want this, so it seems like a natural fit for astro-image-display-api itself.

One generalization is needed: the copies in astrowidgets hard-code `ImageViewerLogic` as the docstring source. An upstream version could either take the interface class as an argument (`@docs_from_interface(ImageViewerLogic)`) or walk the decorated class's MRO to find the method being overridden.

If this lands here, astrowidgets can drop both local copies.

*This issue was written by Claude at Matt Craig's direction.*

mwcraig:
**Stacked PR**: this branch is stacked on #116 → #117 → #118; only the last commit ("Add template-method rendering hooks to ImageViewerLogic") is new here. It will shrink to that commit once those merge.

This turns the public API methods of `ImageViewerLogic` into template methods: they own all state handling and label resolution, then call private no-op rendering hooks that display backends override. Backends therefore never re-implement the public methods, never resolve labels themselves, and never see `None` or `"*"` labels.

## Hook interface

| Hook | Called from | Notes |
|---|---|---|
| `_render_image(label)` | `load_image` | after all state for the new image is stored |
| `_apply_cuts(label)` | `set_cuts`, `load_image` | only when `label` is displayed |
| `_apply_stretch(label)` | `set_stretch`, `load_image` | only when `label` is displayed |
| `_apply_colormap(label)` | `set_colormap`, `load_image` | only when `label` is displayed |
| `_apply_viewport(label)` | `set_viewport`, `load_image` | only when `label` is displayed |
| `_draw_catalog(label)` | `load_catalog`, `set_catalog_style` | |
| `_remove_catalog_marks(label)` | `remove_catalog` | template expands `"*"`, one call per removed catalog |
| `_batch_update()` | wraps the `load_image` hook sequence | context manager, default `contextlib.nullcontext()` |

All hooks receive already-resolved labels and are no-ops in `ImageViewerLogic`, so it remains the valid headless reference implementation and still passes `ImageAPITest` unchanged.

## Displayed-image tracking

A private plural tuple `_displayed_image_labels` is maintained by `load_image` with replace semantics (`(new_label,)`); the `_apply_*` hooks fire only for member labels. This addresses #114 as a private implementation — promotion to the Protocol is deliberately deferred and still open for discussion there.

## Carry-forward on load

`load_image` now captures the displayed image's cuts/stretch/colormap before replacing it and applies them to the newly loaded image in place of the defaults, so e.g. blinking through an image sequence keeps a consistent scaling. This is the behavior the bqplot backend already had; it is now the upstream default, documented in the `load_image` docstring and covered by new behavior tests.

## Also

- `docs_from_super_if_missing` moved here from the astrowidgets backends and exported publicly (`ImageViewerLogic` and the decorator are now importable from the package root), so backends can drop their local copies. Closes #115.
- `api_test.py` gains hook-order, hook-gating, `"*"`-expansion and label-resolution tests via a small recording subclass of the viewer under test, plus the carry-forward tests. 84 tests pass (77 existing + 7 new).
- No Protocol member was added; `interface_definition.py` is untouched.

Closes #115. Addresses #114 (private implementation; Protocol promotion still open).

*This PR was written by Claude at Matt Craig's direction.*


mwcraig:
TODO before (or as part of) this PR merging: now that `_displayed_image_labels` exists, `_catalog_conversion_wcs` in `image_viewer_logic.py` should consult it. #118 left the multi-image case as "ambiguous → raise `ValueError` when a conversion is required, otherwise skip the sky/pixel enrichment," with a note that the rule could tighten once the displayed-image concept landed. That concept lands here, but the tightening was never applied — `_catalog_conversion_wcs` still matches on `len(self._images)` only.

Planned change: in the multiple-images arm, if exactly one image is currently displayed, use that image's WCS instead of raising/skipping; keep the current raise/skip behavior as the fallback when the displayed image is also ambiguous. Practical impact today: loading a sky-coordinate catalog while two WCS images are loaded raises, even though exactly one image is visible and the right WCS is unambiguous.

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/__init__.py
```diff
@@ -7,4 +7,5 @@
 except ImportError:
     __version__ = ""
 
+from .image_viewer_logic import *  # noqa: F403
```
mwcraig: Why is this added to the namespace?

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -302,26 +438,93 @@ def load_image(
         image_label: str | None = None,
         **kwargs,  # noqa: ARG002
     ) -> None:
-        image_label = self._resolve_image_label(image_label, allow_new=True)
+        """
```
mwcraig: This PR should NOT override the docstrings that come from the API definition.

src/astro_image_display_api/api_test.py
```diff
@@ -1353,6 +1354,195 @@ def test_all_methods_accept_additional_kwargs(self, data, catalog, tmp_path):
...
+            the list that the hook calls are recorded in as
+            ``(hook_name, label)`` tuples.
+        """
+        calls = []
+        base_class = self.image_widget_class
+
+        class RecordingViewer(base_class):
+            def _render_image(self, image_label):
+                calls.append(("render_image", image_label))
+                super()._render_image(image_label)
```
mwcraig: Why are the super calls needed here?

src/astro_image_display_api/api_test.py
```diff
@@ -1353,6 +1354,195 @@ def test_all_methods_accept_additional_kwargs(self, data, catalog, tmp_path):
             f"{'\n\t'.join(failed_methods)}"
         )
 
+    # Tests of the template-method hooks: the public API methods own all
+    # state handling and label resolution, then call private no-op hooks
+    # (with already-resolved labels) that backends override to update
+    # their display.
+
+    def _make_recording_viewer(self):
```
mwcraig: None of the tests for the hooks belong in `api_test.py`. Implementations of the API do NOT have to adopt the specific hook mechanism in ImageViewerLogic. They should instead be inside a test file in the tests directory, and should be separate from the existing test that enforces the API contract.

src/astro_image_display_api/__init__.py
mwcraig: Reverted in d59691b — nothing in this repo needs the root import, and the astrowidgets backends already import from `astro_image_display_api.image_viewer_logic` directly, which also works for `docs_from_super_if_missing` once they drop their local copies.

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Removed in d59691b — `load_image` again takes its docstring from the API definition via `docs_from_interface`. The carry-forward behavior is still described in inline comments and enforced by tests; if it should be user-visible documentation, the interface docstring batch in #120 would be the natural place for it.

src/astro_image_display_api/api_test.py
mwcraig: They were defensive: in `api_test.py` the recording subclass wrapped whatever `image_widget_class` a backend put under test, and a real backend's hooks do actual display work, so skipping super() would have changed behavior. Now that the tests target `ImageViewerLogic` directly (d59691b), whose hooks are no-ops, the super() calls are dropped.

src/astro_image_display_api/api_test.py
mwcraig: Moved in d59691b to a new standalone `tests/test_rendering_hooks.py`, separate from the API-contract tests, and `api_test.py` is restored to contract-only. I also moved the two carry-forward tests (`test_load_image_carries_forward_display_settings`, `test_load_image_carry_forward_uses_displayed_image`) even though they only use public API: with the `load_image` docstring override gone, carry-forward is not part of the documented contract, so requiring it of every implementation seemed wrong. Easy to move those two back if you do want carry-forward to be contract-mandated.

*These replies were written by Claude at Matt Craig's direction.*

tests/test_image_viewer_logic_implementation.py
```diff
@@ -0,0 +1,222 @@
...
+    calls.clear()
+
+    viewer.remove_catalog(catalog_label="*")
+
+    removed = [label for name, label in calls if name == "remove_catalog_marks"]
+    assert sorted(removed) == ["cat1", "cat2"]
+    assert viewer.catalog_labels == ()
+
+
+def test_load_image_carries_forward_display_settings(data):
```
mwcraig: Given that this test and the following one are in here, the file needs a different name. Maybe test_image_viewer_logic_implementation?

mwcraig: Renamed to `tests/test_image_viewer_logic_implementation.py` in d6292f5, with the header comment updated to say it covers ImageViewerLogic implementation behavior generally (hooks and carry-forward), not just the hook mechanism.

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -304,24 +440,61 @@ def load_image(
...
+                carried_stretch = displayed.stretch
+                carried_colormap = displayed.colormap
+                break
+
+        with self._batch_update():
+            # Nothing is displayed while the new image's state is being set
+            # up, so the set_* calls made during initialization below do not
+            # fire any of the _apply_* hooks; the hooks are called once, in a
+            # fixed order, at the end of this method.
+            self._displayed_image_labels = ()
```
mwcraig, line 462: **Bug: a failed `load_image` permanently desyncs the displayed-image tracking.**

`_displayed_image_labels` is cleared here, before loading, and never restored if a `_load_*` helper raises (unreadable FITS file, the `NotImplementedError` from `_load_asdf`, ...). After the exception the backend is still displaying the old image, but the template no longer knows it, so every subsequent `set_cuts` / `set_stretch` / `set_colormap` / `set_viewport` on that image stores state without firing its `_apply_*` hook — the screen silently stops tracking the API until the next *successful* `load_image`.

Repro on this branch:

```python
import numpy as np
from astro_image_display_api.image_viewer_logic import ImageViewerLogic

calls = []
class Recording(ImageViewerLogic):
    def _apply_cuts(self, image_label):
        calls.append(image_label)

viewer = Recording()
viewer.load_image(np.ones((5, 5)), image_label="first")
try:
    viewer.load_image("nope.asdf", image_label="second")  # raises NotImplementedError
except NotImplementedError:
    pass

viewer.set_cuts((10, 100), image_label="first")
print(calls)                            # [] -- hook never fired for the still-displayed image
print(viewer._displayed_image_labels)   # ()
print(viewer.image_labels)              # ('first', 'second') -- 'second' is an empty husk
```

The last line shows the related pre-existing wart: `self._images[image_label] = ViewportInfo()` also runs before loading, so a failed load leaves (or clobbers a good image with) an entry whose `data` is `None`. That predates this PR, but it becomes more consequential now that this method is the template every backend relies on. A `try/except` that restores the previous `_displayed_image_labels` and the previous `self._images[image_label]` entry (or deletes the entry if the label was new) before re-raising would fix both.

*This comment was written by Claude (the AI assistant) at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -304,24 +440,61 @@ def load_image(
...
-        else:
-            # Assume it is a 2D array
-            self._load_array(file, image_label)
+                # Assume it is a 2D array
+                self._load_array(file, image_label)
+
+            # Store the carried settings, overriding the defaults set while
+            # loading.
+            if carried_cuts is not None:
+                self._images[image_label].cuts = carried_cuts
```
mwcraig: **Bug: carried settings are shared by reference, not copied — mutating a stretch leaks across images.**

The comment in `test_load_image_carries_forward_display_settings` says "The carried settings are per-image copies of the state", but this code assigns the same objects to the new image. Astropy stretch objects are mutable, so:

```python
import numpy as np
from astropy.visualization import AsinhStretch
from astro_image_display_api.image_viewer_logic import ImageViewerLogic

viewer = ImageViewerLogic()
viewer.load_image(np.ones((5, 5)), image_label="first")
stretch = AsinhStretch(0.1)
viewer.set_stretch(stretch, image_label="first")
viewer.load_image(np.ones((5, 5)) * 2, image_label="second")

print(viewer.get_stretch(image_label="second") is stretch)  # True -- same object
stretch.a = 0.9
print(viewer.get_stretch(image_label="first").a)            # 0.9 -- both images changed
```

The existing test only passes because it *replaces* the second image's cuts with a brand-new object instead of mutating one. Wrapping the three carried values in `copy.copy()` (the module already imports `copy`) would make the behavior match the test's stated intent.

*This comment was written by Claude (the AI assistant) at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -304,24 +440,61 @@ def load_image(
...
-            if isinstance(file, str):
-                is_asdf = file.endswith(".asdf")
-            else:
-                is_asdf = file.suffix == ".asdf"
-            if is_asdf:
-                self._load_asdf(file, image_label)
+        # Carry forward the display settings of the displayed image that is
+        # being replaced, so that the new image appears with the same
+        # cuts/stretch/colormap the user was just looking at.
+        carried_cuts = carried_stretch = carried_colormap = None
```
mwcraig: **Behavioral inconsistency: carry-forward tramples explicit per-label settings, and the "fresh entry" comment below is now wrong.**

Because the carried settings always come from the *displayed* image, reloading data under an existing, currently-not-displayed label silently discards that label's user-chosen settings:

```python
viewer.load_image(d1, image_label="a"); viewer.set_colormap("viridis", image_label="a")
viewer.load_image(d2, image_label="b"); viewer.set_colormap("plasma", image_label="b")

viewer.load_image(d3, image_label="a")           # user refreshes the data for "a"
print(viewer.get_colormap(image_label="a"))      # 'plasma' -- the viridis they set on "a" is gone
```

(Verified on this branch.) Carrying settings to a *new* label is clearly right for the blinking use case; overwriting settings the user explicitly attached to an existing label looks more like data loss. One option: carry forward only when the target label is new.

Relatedly, the comment kept just below — "Start from a fresh entry so that no state from a previous image with this label carries over" — is no longer true: after the first load, no image ever starts from fresh default cuts/stretch/colormap; they always come from the displayed image. Whatever is decided about the behavior, that comment should spell out exactly which state is fresh (data, WCS, viewport) and which is carried.

*This comment was written by Claude (the AI assistant) at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -814,3 +989,33 @@ def get_viewport(
                 fov = viewport.fov
 
         return dict(center=center, fov=fov, image_label=image_label)
+
+
+def docs_from_super_if_missing(cls):
```
mwcraig, line 1014: **Design: near-verbatim duplicate of `docs_from_interface`** — the two functions differ only in the source class, which invites drift (this PR already had to touch both to add the keep-existing-docstring guard). A single parametrized helper keeps them in lockstep:

```python
def _copy_missing_docs(cls, source):
    for name, member in cls.__dict__.items():
        if not name.startswith("_") and not member.__doc__:
            src = getattr(source, name, None)
            if src is not None:
                member.__doc__ = src.__doc__
    return cls

def docs_from_interface(cls):
    return _copy_missing_docs(cls, ImageViewerInterface)

def docs_from_super_if_missing(cls):
    return _copy_missing_docs(cls, ImageViewerLogic)
```

Two smaller points: (a) the name says "super" but the implementation hardcodes `ImageViewerLogic` rather than walking the decorated class's MRO, so a backend with an intermediate base class would not pick up that base's docstrings; (b) despite the PR description, this is not importable from the package root — details in the review summary.

*This comment was written by Claude (the AI assistant) at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -618,8 +784,13 @@ def remove_catalog(
         elif catalog_label == "*":
             # If the user wants to remove all catalogs, we reset the
             # catalogs dictionary to an empty one, which is exactly the
-            # state a fresh viewer starts in.
+            # state a fresh viewer starts in. The "*" is expanded here,
+            # so the _remove_catalog_marks hook is called once per catalog
+            # and never sees the "*" itself.
+            removed_labels = tuple(self._catalogs)
             self._catalogs = {}
+            for removed_label in removed_labels:
```
mwcraig: **Minor design: this loop can fire many `_remove_catalog_marks` calls with no `_batch_update` around them**, unlike the hook burst in `load_image`. If batching exists to let backends suppress intermediate redraws, `remove_catalog("*")` with many catalogs is the same situation. Consider wrapping this loop in `self._batch_update()` — or documenting on `_batch_update` that it is a `load_image`-only affordance.

*This comment was written by Claude (the AI assistant) at Matt Craig's direction.*

mwcraig (commented):
## Review summary

I went through this assuming there were bugs to find; below is what I could actually verify by running code against this branch (d6292f5). Reproductions are in the inline comments; items without a diff line to attach to are listed here.

**Confirmed bugs (see inline comments):**
1. A failed `load_image` clears `_displayed_image_labels` and never restores it, so the `_apply_*` hooks silently stop firing for the image the backend is still displaying.
2. Carry-forward stores shared references, not copies — mutating a carried stretch changes every image that inherited it, contradicting the new test's own comment.

**Behavioral inconsistency (inline):** carry-forward overwrites explicit per-label settings when reloading an existing label, and the retained "fresh entry / no state carries over" comment is now inaccurate.

**PR description vs. implementation** (all verified on the branch):
- *"`ImageViewerLogic` and the decorator are now importable from the package root"* — they are not. `__init__.py` still only does `from .interface_definition import *`, and that module's `__all__` is `["ImageViewerInterface"]`. On this branch `from astro_image_display_api import docs_from_super_if_missing` raises `ImportError`, so #115 would not actually be closed by merging this.
- *"documented in the `load_image` docstring"* — the carry-forward behavior is not in `load_image`'s docstring. The class method has no docstring of its own, so it inherits the interface docstring, whose Notes section mentions only the viewport.
- *"`api_test.py` gains hook-order … tests via a small recording subclass of the viewer under test"* — `api_test.py` is untouched; the tests live in the new `tests/test_image_viewer_logic_implementation.py` and exercise `ImageViewerLogic` only. That scoping is defensible (the test header explicitly says implementations need not adopt the template), but note the consequence: the carry-forward is a user-observable behavior change of `load_image`, and nothing in the compliance suite requires a backend to preserve it, so backends that re-implement `load_image` can silently diverge. Worth an explicit decision: promote it to the interface docs + `ImageAPITest`, or scope the PR description down to match.
- Test count nit: the description says 84 tests pass; the branch has 87.

**Design questions with no inline anchor:**
- The stated motivation is blinking through an image sequence, but the **viewport is not carried forward** — every `load_image` re-centers and re-zooms via `_initialize_image_viewport_stretch_cuts`. When blinking aligned images, preserving pan/zoom seems at least as important as preserving the scaling. Is dropping it deliberate (bqplot parity), or an oversight?
- There is **no hook for an image ceasing to be displayed**. With single-image replace semantics `_render_image(new_label)` implies it, but the tuple exists specifically to leave room for multi-image display, and at that point backends will need to know which labels were un-displayed. Might be worth either a `_clear_image(label)` hook now or an explicit note in the `_render_image` docstring that replace-implies-clear is part of the contract.

*This review was written by Claude (the AI assistant) at Matt Craig's direction. No code was changed; comments only.*

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -304,24 +440,61 @@ def load_image(
...
+        # being replaced, so that the new image appears with the same
+        # cuts/stretch/colormap the user was just looking at.
+        carried_cuts = carried_stretch = carried_colormap = None
+        for displayed_label in self._displayed_image_labels:
+            displayed = self._images.get(displayed_label)
+            if displayed is not None:
+                carried_cuts = displayed.cuts
+                carried_stretch = displayed.stretch
+                carried_colormap = displayed.colormap
+                break
```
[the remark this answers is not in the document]
mwcraig: Confirmed — the carried values are assigned by reference, and astropy stretch objects are mutable, so an in-place mutation leaks across images. The same issue was flagged with a runnable repro in the thread at https://github.com/astropy/astro-image-display-api/pull/119#discussion_r3607506767; the existing carry-forward test only passes because it replaces the cuts object rather than mutating it. The suggested fix is right: `copy.copy()` the three carried values when capturing them (the module already imports `copy`).

*This reply was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -304,24 +440,61 @@ def load_image(
...
+        with self._batch_update():
+            # Nothing is displayed while the new image's state is being set
+            # up, so the set_* calls made during initialization below do not
+            # fire any of the _apply_* hooks; the hooks are called once, in a
+            # fixed order, at the end of this method.
+            self._displayed_image_labels = ()
+
+            # Start from a fresh entry so that no state from a previous image
+            # with this label carries over.
+            self._images[image_label] = ViewportInfo()
```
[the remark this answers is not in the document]
mwcraig: Confirmed — a `_load_*` failure (unreadable FITS, the `NotImplementedError` from `_load_asdf`) leaves `_displayed_image_labels` empty while the backend is still displaying the old image, so subsequent `set_cuts`/`set_stretch`/`set_colormap`/`set_viewport` calls store state without firing their `_apply_*` hooks until the next successful `load_image`. There is a runnable repro in the thread at https://github.com/astropy/astro-image-display-api/pull/119#discussion_r3607506762, which also notes the related problem that `self._images[image_label] = ViewportInfo()` runs before loading too, so a failed load leaves an empty-husk entry (or clobbers a good image under that label). A `try/except` that restores both the previous `_displayed_image_labels` and the previous `_images[image_label]` entry (deleting it if the label was new) before re-raising would fix both at once, which goes a bit further than the restore suggested here.

*This reply was written by Claude at Matt Craig's direction.*

tests/test_image_viewer_logic_implementation.py
```diff
@@ -0,0 +1,224 @@
...
+
+@pytest.fixture
+def catalog():
+    rng = np.random.default_rng(45328975)
+    return Table(
+        dict(
+            x=rng.uniform(0, IMAGE_SHAPE[0], size=10),
+            y=rng.uniform(0, IMAGE_SHAPE[1], size=10),
+        )
+    )
```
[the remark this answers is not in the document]
mwcraig: Correct on the convention — with `IMAGE_SHAPE = (100, 150)` as `(height, width)`, `x` should span `[0, IMAGE_SHAPE[1])` and `y` should span `[0, IMAGE_SHAPE[0])`, and this fixture has them swapped. It is harmless in practice here: `ImageViewerLogic` never validates catalog coordinates against the image shape, and these tests only exercise hook dispatch and label bookkeeping. Worth noting the swap was inherited: the pre-existing `catalog` fixture in `api_test.py` (lines 54–55) generates its coordinates the same way, so if this gets fixed, both fixtures should be fixed together for consistency.

*This reply was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: I think I want to change this behavior. Instead of carry-forward, loading an image with an image label that already exists would use that image label's cuts/stretch/colormap (not viewport necessarily). Loading an image with a new image label gets the default values.

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Sure, do that.

tests/test_image_viewer_logic_implementation.py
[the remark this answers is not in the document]
mwcraig: Fix them both

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Implemented in bbf8a8a: carry-forward is gone. `load_image` under an existing label keeps that label's cuts/stretch/colormap (the viewport is reset to fit the new data); a new label starts from the default settings. The implementation tests were rewritten to pin both cases, and the `load_image` docstring added on the docs branch now describes this behavior.

*This reply was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Fixed in bbf8a8a — a failed load now restores `_displayed_image_labels` and the previous `_images` entry (or deletes the fresh one when the label was new) before re-raising, so the hooks keep firing for the image still on screen and no empty-husk entry is left. Two regression tests added.

*This reply was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Superseded by the behavior change in bbf8a8a: settings are no longer shared across labels at all, so there is no reference-aliasing left to copy around.

*This reply was written by Claude at Matt Craig's direction.*

tests/test_image_viewer_logic_implementation.py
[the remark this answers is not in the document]
mwcraig: Both fixed in bbf8a8a — the fixtures in `api_test.py` and `tests/test_image_viewer_logic_implementation.py` now draw x from the width and y from the height.

*This reply was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Done in bbf8a8a — both decorators now delegate to a shared `_copy_missing_docs(cls, source)` helper.

*This reply was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Done in bbf8a8a — the `remove_catalog("*")` loop now runs inside `_batch_update`, and the star-expansion test asserts the hooks fire within one batch.

*This reply was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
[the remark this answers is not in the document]
mwcraig: Fixed in bbf8a8a; see the sibling thread for details.

*This reply was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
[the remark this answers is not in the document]
mwcraig: Superseded by bbf8a8a — settings are no longer carried across labels, so no shared references remain.

*This reply was written by Claude at Matt Craig's direction.*

mwcraig:
The `_catalog_conversion_wcs` tightening described in the TODO above is implemented in 301b797: with several images loaded, if exactly one image is currently displayed its WCS is used for catalog pixel/sky conversion; the raise/skip behavior remains as the fallback when the displayed image is also ambiguous. The two `api_test` contract tests that asserted the old raise/skip behavior now assert that the displayed image's WCS is the one used.

*This comment was written by Claude at Matt Craig's direction.*

PR #120, opened 2026-07-17
Docs and compliance-test cleanup: docstring drift, colormap policy, parameter-name conformance

### LINKED ISSUE #98: api_test: tests that contradict the interface contract and would fail compliant backends
1. `test_stretch` and `test_cuts` call `get_stretch`/`set_stretch`/`set_cuts` on a viewer with **no image loaded** and expect success ([api_test.py#L724-L743](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L724-L743)). The interface documents `ValueError` for labels that don't correspond to a loaded image, so a *compliant* backend fails these tests; they only pass on the reference implementation because of M2.
2. `test_load_multiple_catalogs` requires the exact error text `"Multiple catalog styles defined."` from `get_catalog()` ([api_test.py#L568-L570](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L568-L570)) — a message that (a) is an implementation detail of the reference class and (b) is wrong on its face (getting catalog *data*, message talks about *styles* — see M9).
3. `test_stretch` uses `is`/`is not` identity checks ([api_test.py#L731](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L731), [#L735](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L735)) — backends returning an equivalent-but-new stretch object fail.

Since api_test.py is the compliance suite every backend must pass, these effectively redefine the API away from what interface_definition.py documents.

---

### LINKED ISSUE #99: api_test: weak assertions and coverage gaps that let broken backends pass
- `test_load` performs no assertions at all after `load_image` ([api_test.py#L86-L100](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L86-L100)).
- `test_get_image` only asserts the result `is not None`; it never compares retrieved data to what was loaded ([api_test.py#L874-L887](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L874-L887)).
- The WCS fixture has `|cdelt[0]| == |cdelt[1]|` ([api_test.py#L40-L41](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L40-L41)), so no test can detect axis-swapped pixel-scale handling (bug H1). Add an anisotropic-WCS + rectangular-image FOV test and a full sky→pixel→sky round trip that compares against the original *pixel* values (the current `test_viewport_round_trips` compares sky-to-sky only, [api_test.py#L399-L417](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L399-L417)).
- `get_catalog` with non-default `x_colname`/`y_colname` is never tested (would have caught H2).
- Loading/retrieving *unlabeled* images and catalogs alongside labeled ones is never tested (would have caught H4).
- The `catalog` fixture draws x from `[0, shape[0])` and y from `[0, shape[1])` — ranges swapped relative to (height, width); harmless today but misleading ([api_test.py#L53-L55](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L53-L55)).

---

### LINKED ISSUE #102: `get_cuts` annotated `-> tuple`, contract says `BaseInterval`
[image_viewer_logic.py#L168-L172](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L168-L172) vs [interface_definition.py#L139](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L139). The implementation always stores an interval (tuples are converted by `set_cuts`), so the annotation is simply wrong; it can also return `None` before anything is loaded (see M2).

---

### LINKED ISSUE #103: Docstring drift: `set_catalog_style` missing `catalog_label`; stale `remove_catalog` docstring in the implementation
- The interface `set_catalog_style` docstring documents `shape`, `color`, `size`, `**kwargs` but not the first parameter `catalog_label` ([interface_definition.py#L393-L434](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L393-L434)).
- `ImageViewerLogic.remove_catalog` carries a stale docstring documenting a nonexistent `marker_name` parameter ([image_viewer_logic.py#L538-L546](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L538-L546)). It's masked at runtime by the `docs_from_interface` decorator, which overwrites implementation docstrings — given that, in-source docstrings in the implementation that duplicate (old versions of) interface docs should just be removed.

---

### LINKED ISSUE #104: `docs/testing_implementation.rst` references a nonexistent class `ImageDisplayInterface`
[docs/testing_implementation.rst#L9-L11](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/docs/testing_implementation.rst#L9-L11) tells implementers to check `isinstance(your_instance, ImageDisplayInterface)`; the class is `ImageViewerInterface`. Anyone following the docs gets a `NameError`.

---

### LINKED ISSUE #105: Duplicate test: `test_every_method_attribute_has_docstring` exists in both `api_test.py` and `tests/`
Nearly verbatim copies: [api_test.py#L983-L1003](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L983-L1003) and [tests/test_astro_image_display_api.py#L77-L97](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/tests/test_astro_image_display_api.py#L77-L97) (one checks the instance, the other the protocol — if both are wanted, the shared body should still be one function). Given the redundancy-removal priority, keep one.

---

### LINKED ISSUE #107: `get_catalog_style`: tests require an undocumented `catalog_label` key and a default style before any catalog is loaded
The compliance suite requires two behaviors of `get_catalog_style` that the interface docstring does not document — a backend that implements exactly what the docstring says will fail `api_test.py`.

1. **Every return must include a `catalog_label` key.** The documented Returns section says only "The style of the markers" ([interface_definition.py#L453-L456](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L453-L456)), but `test_set_get_catalog_style_with_single_label` asserts equality against a dict that *includes* `catalog_label` ([api_test.py#L450-L460](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L450-L460)), and `test_set_get_catalog_style_preserves_extra_keywords` has to `del retrieved_style["catalog_label"]` before comparing ([api_test.py#L488-L490](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L488-L490)). The reference implementation injects the key unconditionally ([image_viewer_logic.py#L236-L238](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L236-L238)).

2. **With no catalogs loaded, a full default style must be returned.** `test_set_get_catalog_style_no_labels` calls `get_catalog_style()` on a viewer with no catalog and requires `color`/`shape`/`size` keys ([api_test.py#L427-L434](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L427-L434)). The docstring says nothing about this case — and its only statement about labels that don't correspond to a loaded catalog promises an *empty* dict ([interface_definition.py#L443-L448](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L443-L448)), which suggests the opposite.

Whichever behavior is intended, docstring and test should say the same thing: either document the `catalog_label` key and the default-style guarantee in the interface, or drop them from the tests.

Related but distinct: #95 (the empty-dict promise is not implemented, plus `defaultdict` pollution), #96 (default style lost after `remove_catalog('*')`), #103 (`catalog_label` missing from the `set_catalog_style` parameter docs).

---
*From a code review of `main` @ [`7f8e611`](https://github.com/astropy/astro-image-display-api/commit/7f8e611994deda64244ca49f990b5f8a1cb0ed69); this second pass focused on the API contract (interface docstrings vs `api_test.py` vs design).*

### LINKED ISSUE #108: `load_catalog` docstring: "if given, `use_skycoord` is ignored" is illogical and contradicted by the tests
The `load_catalog` docstring says of `skycoord_colname` ([interface_definition.py#L361-L364](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L361-L364)):

> The name of the column containing the sky coordinates. If given, the ``use_skycoord`` parameter is ignored.

Two problems:

1. **Illogical as written.** `skycoord_colname` has a default (`'coord'`), so it is always "given" — read literally, `use_skycoord` would never do anything. Presumably the intent was "if the table *contains* the sky-coordinate column", but that isn't what it says.

2. **Contradicted by the tests and the reference implementation.** `test_load_catalog_with_skycoord_no_wcs` loads a table whose *only* position column is `coord` (with no WCS and the default `use_skycoord=False`) and requires a `ValueError` about pixel coordinates ([api_test.py#L599-L609](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L599-L609)). If `use_skycoord` were ignored when a sky-coordinate column is present, this load would have to succeed. The reference implementation likewise branches on `use_skycoord` after checking the columns ([image_viewer_logic.py#L499-L509](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L499-L509)) — `use_skycoord=True` with the same table (plus a WCS or the coord column) takes a different path than `use_skycoord=False`.

The docstring sentence should be rewritten to describe the actual (test-enforced) semantics: `use_skycoord` decides whether positions are taken from the sky-coordinate column, and errors are raised when the requested mode is impossible with the available columns/WCS. The two `ValueError` messages the tests require ("Cannot use pixel coordinates without pixel columns…", "Cannot use sky coordinates without…") are also absent from the docstring's Raises section ([interface_definition.py#L379-L384](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L379-L384)), which currently only mentions missing required columns and label ambiguity.

Related: #92 covers unlabeled-catalog storage and the "unique name will be generated" promise; it does not touch the `use_skycoord`/`skycoord_colname` semantics.

---
*From a code review of `main` @ [`7f8e611`](https://github.com/astropy/astro-image-display-api/commit/7f8e611994deda64244ca49f990b5f8a1cb0ed69); this second pass focused on the API contract (interface docstrings vs `api_test.py` vs design).*

### LINKED ISSUE #109: Colormap contract: invalid-name behavior is self-contradictory and untested; no default colormap after load, unlike stretch/cuts
The colormap half of the API is the least specified of the per-image display settings, and what is specified is self-contradictory.

1. **Invalid-name behavior contradicts itself within one docstring.** The `map_name` parameter text says backends that don't support a colormap "should handle errors gracefully" ([interface_definition.py#L242-L246](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L242-L246)), while the Raises section promises `ValueError` "If the ``map_name`` is not a valid colormap name" ([interface_definition.py#L255-L259](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L255-L259)). A compliant backend cannot both raise and tolerate. Neither behavior is enforced: the reference implementation stores any value without validation ([image_viewer_logic.py#L202-L213](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L202-L213)) and no test covers an invalid name.

2. **No initialized-colormap-after-load guarantee, unlike stretch and cuts.** `get_colormap` is annotated `-> str` and documents its return as "The name of the current colormap" ([interface_definition.py#L270-L287](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L270-L287)), but on the reference implementation a freshly loaded image has `colormap=None` ([image_viewer_logic.py#L52](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L52)) — `load_image` initializes stretch and cuts but never a colormap. Stretch and cuts have an explicit post-load guarantee with a test (`test_stretch_cuts_are_set_after_loading_image`, [api_test.py#L770-L780](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L770-L780)); colormap has neither the guarantee nor the test. Either the same guarantee should apply (a backend-chosen default colormap after load, tested the same way) or the docstring should document that `get_colormap` may return `None`/a sentinel before one is set.

Related: #94 shows `set_colormap` succeeding on an *empty* viewer (dead label checks); #102 notes `get_cuts` can return `None` pre-load. Neither covers the invalid-name contradiction or the missing post-load guarantee.

---
*From a code review of `main` @ [`7f8e611`](https://github.com/astropy/astro-image-display-api/commit/7f8e611994deda64244ca49f990b5f8a1cb0ed69); this second pass focused on the API contract (interface docstrings vs `api_test.py` vs design).*

### LINKED ISSUE #110: Viewport contract gaps: `image_label` return documented as str but must be None; undocumented "WCS is not set" error; three-way error taxonomy for bad center/fov
Three places where the viewport contract in the interface docstrings and the behavior enforced by `api_test.py` disagree. Backends written from the docstrings alone will fail the compliance suite.

1. **`get_viewport`'s returned `image_label` is documented as a string but must be `None` for unlabeled images.** The Returns section says "'image_label' is a string representing the label of the image" ([interface_definition.py#L643-L651](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L643-L651)), but `test_viewport_is_defined_after_loading_image` loads an image without a label and asserts `vport["image_label"] is None` ([api_test.py#L243-L244](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L243-L244)).

2. **A test-required `ValueError` is missing from `get_viewport`'s Raises section.** `get_viewport(sky_or_pixel="sky")` on an image with no WCS must raise `ValueError` matching "WCS is not set" ([api_test.py#L219-L221](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L219-L221)), but the docstring's Raises section lists only invalid `sky_or_pixel` values and label problems ([interface_definition.py#L653-L658](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L653-L658)).

3. **Three different error styles for bad `center`/`fov` in `set_viewport`, all hard-coded in the tests.** The suite requires: `u.UnitTypeError` "Incorrect unit for fov" for a non-angular Quantity ([api_test.py#L168-L169](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L168-L169)); `TypeError` "Invalid value for fov"/"Invalid value for center" for wrong types when a WCS is present ([api_test.py#L171-L177](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L171-L177)); and *differently worded* `TypeError`s "Center must be a tuple"/"FOV must be a float" for sky values when no WCS is present ([api_test.py#L209-L217](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L209-L217)). The docstring lumps all of these into one generic `TypeError` clause plus `UnitTypeError` ([interface_definition.py#L594-L607](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L594-L607)), so an implementer has no way to know the exact messages the suite pattern-matches. Either the docstrings should specify the required message fragments, or the tests should match only on exception type.

Related: #93 (which WCS decides the sky/pixel default — an implementation bug, not a docstring gap) and #98 (other cases where api_test contradicts the contract).

---
*From a code review of `main` @ [`7f8e611`](https://github.com/astropy/astro-image-display-api/commit/7f8e611994deda64244ca49f990b5f8a1cb0ed69); this second pass focused on the API contract (interface docstrings vs `api_test.py` vs design).*

### LINKED ISSUE #111: Reference implementation parameter names diverge from the interface (`data`/`file`, `cuts`/`value`, `stretch`/`value`); keyword-name conformance is untested
The interface defines keyword parameter names that are part of the documented API, but the reference implementation uses different names for three of them, and nothing in the compliance suite checks names — so the divergence is invisible to the tests.

| Method | Interface name | `ImageViewerLogic` name |
| --- | --- | --- |
| `load_image` | `data` ([interface_definition.py#L22](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L22)) | `file` ([image_viewer_logic.py#L294-L297](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L294-L297)) |
| `set_cuts` | `cuts` ([interface_definition.py#L98-L101](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L98-L101)) | `value` ([image_viewer_logic.py#L180-L182](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L180-L182)) |
| `set_stretch` | `stretch` ([interface_definition.py#L170-L175](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L170-L175)) | `value` ([image_viewer_logic.py#L150-L152](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L150-L152)) |

Consequence: code written against the documented API fails on the reference implementation —

```python
v = ImageViewerLogic()
v.load_image(data=np.zeros((10, 10)))
# TypeError: load_image() got an unexpected keyword argument 'data'
v.set_cuts(cuts=(0, 1))
# TypeError: set_cuts() got an unexpected keyword argument 'cuts'
```

This slips through because `typing.runtime_checkable` protocols only check attribute *existence*, not signatures, and every call in `api_test.py` passes these arguments positionally.

Suggested direction: rename the implementation parameters to match the interface, and add a conformance test that calls each protocol method using exactly the keyword names from the interface signature (e.g. via `inspect.signature(ImageViewerInterface.<method>)`), so future backends can't drift either.

Related: #99 lists other api_test coverage gaps (weak assertions, column-name arguments, unlabeled data), but not keyword-name conformance.

---
*From a code review of `main` @ [`7f8e611`](https://github.com/astropy/astro-image-display-api/commit/7f8e611994deda64244ca49f990b5f8a1cb0ed69); this second pass focused on the API contract (interface docstrings vs `api_test.py` vs design).*

### LINKED ISSUE #112: Interface docstring drift: copy-paste errors and missing Raises sections in `get_image`, `get_stretch`, `get_cuts`, `load_image`, `remove_catalog`
A bundle of small copy-paste and omission errors in the interface docstrings. Individually minor, but these docstrings are the API specification that backend authors implement from (and `docs_from_interface` propagates them verbatim onto implementations).

- **`get_image`**: the `image_label` description reads "The label of the image to **set the cuts** for" ([interface_definition.py#L54-L59](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L54-L59)) — copied from `set_cuts`. The docstring also has no summary line; it opens directly with the Parameters section.
- **`get_stretch`**: the `image_label` description says "get **the cuts** for … the cuts for that image are returned" ([interface_definition.py#L212-L217](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L212-L217)), and there is no Raises section at all — even though the compliance suite requires `ValueError` "Image label ... not found" from `get_stretch` ([api_test.py#L787-L788](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L787-L788)), which sibling getters like `get_cuts` do document.
- **`get_cuts`**: the `kwargs` entry sits under **Returns** instead of Parameters ([interface_definition.py#L150-L156](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L150-L156)), so rendered docs list it as a return value.
- **`load_image`**: no Raises section, and loading with an existing `image_label` silently replaces that image (reference implementation: [image_viewer_logic.py#L302-L305](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L302-L305)) without the docstring saying so — `load_catalog` documents its identical overwrite behavior explicitly ([interface_definition.py#L348-L349](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L348-L349)).
- **`remove_catalog`**: the Raises section promises `TypeError` "If the ``catalog_label`` is not a string or `None`" ([interface_definition.py#L494-L496](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/interface_definition.py#L494-L496)), but the reference implementation raises `TypeError` only for a `list` ([image_viewer_logic.py#L547-L551](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/image_viewer_logic.py#L547-L551)); any other non-string (e.g. an `int`) falls through to a `ValueError`. The test only covers the `list` case ([api_test.py#L689-L698](https://github.com/astropy/astro-image-display-api/blob/7f8e611994deda64244ca49f990b5f8a1cb0ed69/src/astro_image_display_api/api_test.py#L689-L698)).

Related: #103 covers two other drift items (`set_catalog_style` missing its `catalog_label` parameter doc; a stale `remove_catalog` docstring in the implementation), and #101 covers the copy-paste error strings inside the label-resolution helpers. The items above are all distinct from those.

---
*From a code review of `main` @ [`7f8e611`](https://github.com/astropy/astro-image-display-api/commit/7f8e611994deda64244ca49f990b5f8a1cb0ed69); this second pass focused on the API contract (interface docstrings vs `api_test.py` vs design).*

mwcraig:
Docs/test cleanup batch from the July 2026 API-contract review.

> [!NOTE]
> **Stacked PR**: based on #119 (template hooks), which is based on #118 <- #117 <- #116. Only the last two commits (`3d46536`, `92807ea`) are new here; the diff will shrink to those once the others merge.

## Docs batch (commit 1)

- **#104**: `docs/testing_implementation.rst` told implementers to `isinstance`-check against the nonexistent `ImageDisplayInterface`; fixed to `ImageViewerInterface`.
- **#112**: interface docstring drift — `get_image` summary line + copy-pasted `image_label` text, `get_stretch` cuts/stretch copy-paste + missing Raises, `get_cuts` `kwargs` moved from Returns to Parameters, `load_image` now documents that re-using an `image_label` replaces the image, `remove_catalog`'s documented `TypeError` now matches the actual list-only behavior.
- **#102**: `ImageViewerLogic.get_cuts` annotated `-> BaseInterval` (it always returns an interval; the pre-load `None` case was removed by the empty-viewer guards in #116).
- **#103**: `set_catalog_style` now documents its `catalog_label` parameter. The stale `remove_catalog` docstring in the implementation was already removed in the earlier PRs.
- **#107**: `get_catalog_style` Returns now documents the always-present `catalog_label` key and the full default style (with `catalog_label=None`) returned when no catalog is loaded; the stale empty-dict promise was already replaced by the raise-on-unknown-label contract (#113) in #117.
- **#108**: `load_catalog` no longer claims `skycoord_colname` makes `use_skycoord` ignored; the docstring now describes the test-enforced semantics (`use_skycoord` picks the position source, errors when the requested mode is impossible) and documents both required `ValueError`s.
- **#110**: `set_viewport` Raises now spells out the three error styles and their message fragments the compliance suite matches; `get_viewport` documents the "WCS is not set" `ValueError`. Item 1 of the issue (returned `image_label` documented as str but `None` for unlabeled loads) was resolved the other way by #117: unlabeled loads now get generated labels, so the return really is always a string — the docstring now mentions generated labels.
- **#109**: colormap policy (see below).

## Tests + conformance (commit 2)

- **#111**: renamed the `ImageViewerLogic` parameters that diverged from the interface — `load_image(file=` → `data=`, `set_cuts(value=` → `cuts=`, `set_stretch(value=` → `stretch=` — and added `test_parameter_names_match_interface`, which checks by signature inspection that every implementation uses the interface's parameter names in the interface's order (calling by keyword can't catch this because `**kwargs` silently swallows misnamed keywords).
  - **astrowidgets impact**: checked the slimmed `ginga.py`/`bqplot.py` backends (and their tests): neither overrides these three methods nor calls them with the old keyword names anywhere, so the renames are safe. The ginga `set_colormap` override already uses the interface's `map_name`.
- **#98**: `test_cuts` now loads an image before exercising `set_cuts` (the contract only defines behavior for loaded images; `test_stretch` was already fixed in #116-#118), and the `is`/`is not` identity assertions in `test_stretch` are replaced with type checks so backends returning equivalent-but-new stretch objects pass. The wrong-on-its-face `"Multiple catalog styles defined."` message requirement was already fixed in #117 (now `"Multiple catalog labels defined"`).
- **#99**: `test_load` now asserts the image is registered and its data round-trips through `get_image`; `test_get_image` compares retrieved data to what was loaded; `test_viewport_round_trips` completes the pixel→sky→pixel round trip against the original pixel values; the catalog fixture's swapped x/y ranges are fixed. The anisotropic-WCS FOV test, custom-colname `get_catalog` tests, and unlabeled-alongside-labeled tests were already added in #116-#117.
- **#105**: removed the near-verbatim duplicate `test_every_method_attribute_has_docstring` from `tests/`, keeping the copy in the shipped compliance suite (which runs against every implementation, including the reference whose docstrings come from the protocol).

## #109 colormap policy chosen

The issue had no discussion comments, so the policy was chosen to match the direction of the rest of the review (strict `ValueError`s, e.g. #113) and the behavior of the existing backends:

1. **Invalid names raise `ValueError`** — the "should handle errors gracefully" wording is removed; the contract now says a backend must raise `ValueError` for a colormap name it does not support, and the compliance suite tests this with an obviously bogus name (matching loosely on "not a valid", which fits the ginga backend's existing message and matplotlib's own). The reference implementation validates against the matplotlib colormap registry when matplotlib is installed (matplotlib stays out of the core dependencies; it is added to the `test` extra so CI exercises the validation).
2. **No mandated post-load default colormap** — `get_colormap` is documented and annotated `str | None`: `None` means no colormap has been set and the backend's default is in use. This matches both astrowidgets backends (ginga leaves it unset until the user sets one; bqplot reports its own default after first display) and avoids forcing a name on backends whose default has no matplotlib name. Mandating a reference-chosen default instead would have silently disabled the bqplot backend's "first display" default-injection logic.

Closes #98, closes #99, closes #102, closes #103, closes #104, closes #105, closes #107, closes #108, closes #109, closes #110, closes #111, closes #112.

Full suite: 84 tests pass (one test removed as duplicate, one conformance test added).


mwcraig:
TODO for this PR (from [review discussion on #118](https://github.com/astropy/astro-image-display-api/pull/118#discussion_r3606363724)): consolidate the WCS construction in `api_test.py`. The three inline TAN-WCS constructions in `test_fov_anisotropic_pixel_scale`, `test_fov_sky_to_pixel_honors_quantity_unit`, and `test_fov_wcs_with_arcsec_cunit` should reuse the `_make_tan_wcs` helper added in #118, extending it to take a per-axis `cdelt` pair (and `cunit` where needed) instead of the single isotropic `scale`.

*This comment was written by Claude at Matt Craig's direction.*

pyproject.toml
```diff
@@ -35,6 +35,7 @@ docs = [
 ]
 test = [
     "black",
+    "matplotlib",
```
mwcraig, line 38: Why does that need to be installed??

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -195,58 +330,120 @@ def get_stretch(
...
+        if not isinstance(stretch, BaseStretch):
             raise TypeError(
-                f"Stretch option {value} is not valid. Must be an "
+                f"Stretch option {stretch} is not valid. Must be an "
                 "`astropy.visualization` Stretch object."
             )
         image_label = self._resolve_image_label(image_label)
-        self._images[image_label].stretch = value
+        self._images[image_label].stretch = stretch
+        if image_label in self._displayed_image_labels:
```
mwcraig, line 344: Just to make sure I understand, the only thing gated on whether an image is displayed  is actually applying the stretch to the image? So `get_stretch` for the label returns the stretch that was just set here even if the image is not displayed?

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -195,58 +330,120 @@ def get_stretch(
...
+        if map_name not in colormaps:
+            raise ValueError(f"Colormap '{map_name}' is not a valid colormap name.")
 
     def set_colormap(
         self,
         map_name: str,
         image_label: str | None = None,
         **kwargs,  # noqa: ARG002
     ) -> None:
+        """
```
mwcraig: Why is this docstring being overridden?

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -298,30 +497,100 @@ def _resolve_image_label(
 
     def load_image(
         self,
-        file: str | os.PathLike | ArrayLike | NDData,
+        data: str | os.PathLike | ArrayLike | NDData,
         image_label: str | None = None,
         **kwargs,  # noqa: ARG002
     ) -> None:
-        image_label = self._resolve_image_label(image_label, allow_new=True)
+        """
```
mwcraig, line 474: Unless there is some very good reason to we should not be overridding this. What is the reason for overriding the interface docstring?

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -298,30 +497,100 @@ def _resolve_image_label(
...
+                    if is_asdf:
+                        self._load_asdf(data, image_label)
+                    else:
+                        self._load_fits(data, image_label)
+                elif isinstance(data, NDData):
+                    self._load_nddata(data, image_label)
+                else:
+                    # Assume it is a 2D array
+                    self._load_array(data, image_label)
+            except Exception:
```
mwcraig, line 534: Can this be a more specific than `Exception`?

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -482,30 +751,36 @@ def _catalog_conversion_wcs(self, conversion_required: bool) -> WCS | None:
         Returns
         -------
         `astropy.wcs.WCS` or None
-            The WCS of the single loaded image, or None.
```
mwcraig, line 485: Why does get_image override the interface docstring?

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -298,30 +497,100 @@ def _resolve_image_label(
...
+            try:
+                if isinstance(data, str | os.PathLike):
+                    if isinstance(data, str):
+                        is_asdf = data.endswith(".asdf")
+                    else:
+                        is_asdf = data.suffix == ".asdf"
+                    if is_asdf:
+                        self._load_asdf(data, image_label)
+                    else:
+                        self._load_fits(data, image_label)
```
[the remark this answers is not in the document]
mwcraig: Confirmed — this is a real bug, though it predates this PR: the same code is already on `main`, and this PR only renames `file` → `data` inside it, which is why the lines show as changed here. Reproduction against this branch (with `image.fits` a valid FITS file in `tmp`):

```python
import os
from astro_image_display_api import ImageViewerLogic

viewer = ImageViewerLogic()
entry = next(e for e in os.scandir(tmp) if e.name == "image.fits")
isinstance(entry, os.PathLike)   # True
viewer.load_image(entry)
# AttributeError: 'posix.DirEntry' object has no attribute 'suffix'
```

Any `os.PathLike` that is not a `pathlib.PurePath` fails the same way (`os.DirEntry`, custom `__fspath__` classes). Two mitigating details: the failure is caught by `load_image`'s restore-and-reraise handler, so the viewer's state stays intact (verified in the same run), and str/`Path` cover nearly all real callers.

The whole str-vs-PathLike branch can collapse to one line that handles str, bytes-returning `__fspath__`, and every `os.PathLike` uniformly:

```python
is_asdf = os.fsdecode(data).endswith(".asdf")
```

Since the bug is pre-existing rather than introduced by this PR, it has not been folded in without Matt's sign-off — it could go in here as a one-liner or in a follow-up.

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -814,3 +1092,26 @@ def get_viewport(
                 fov = viewport.fov
 
         return dict(center=center, fov=fov, image_label=image_label)
+
+
+def docs_from_super_if_missing(cls):
+    """
+    Class decorator that fills in missing docstrings from the AIDA interface.
```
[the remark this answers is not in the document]
mwcraig: Partially disagree. `docs_from_super_if_missing` is not touched by this PR — it is already on `main`, and this comment is anchored on the pre-merge full-stack diff. Only the one-line summary says "the AIDA interface"; the paragraph directly under it states the actual source correctly:

> Public methods of the decorated class that lack a docstring receive the docstring of the same-named method on `ImageViewerLogic`.

Until this PR the distinction was also invisible in practice: every public docstring on `ImageViewerLogic` was itself inherited from `ImageViewerInterface` via `@docs_from_interface`, so copying from the class *was* copying from the interface.

That said, the underlying concern does become real with this PR: the new `set_colormap`/`load_image` docstrings defined directly on `ImageViewerLogic` mean a backend using the decorator now inherits reference-implementation-specific text (e.g. "validated against the matplotlib colormap registry") that may not describe that backend. That issue is being discussed on the threads for those two docstrings; if the overrides are dropped there, this comment reduces to a one-word wording nit in the summary line.

*This comment was written by Claude at Matt Craig's direction.*

pyproject.toml
mwcraig: It is needed by the new compliance test, not at runtime. `api_test.py::test_set_colormap_errors` now requires every backend to raise `ValueError` for a bogus colormap name (the #109 policy). The reference implementation validates names against the matplotlib registry, but only when matplotlib is importable — matplotlib deliberately stays out of the runtime dependencies. Without it in the `test` extra, the validation silently no-ops and the reference implementation fails its own compliance suite:

```python
import sys
sys.modules["matplotlib"] = None   # simulate matplotlib not installed
import numpy as np
from astro_image_display_api import ImageViewerLogic

viewer = ImageViewerLogic()
viewer.load_image(np.ones((5, 5)), image_label="test")
viewer.set_colormap("this-is-not-a-colormap", image_label="test")  # accepted — no ValueError
```

The honest caveat: this makes the reference implementation compliance-clean only in environments that happen to have matplotlib. The alternatives each trade something else away — a hard matplotlib dependency (heavy for a package this small), skipping the colormap test when matplotlib is absent (weakens the contract for every backend, not just this one), or vendoring a name list (drifts). If the optional-validation compromise feels too clever, the cleanest alternative is making matplotlib a real dependency; that is a call for you to make.

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Yes, exactly. The per-label state is updated unconditionally (`self._images[image_label].stretch = stretch` on the line above); only the rendering hook `_apply_stretch` is gated on the image being displayed. `get_stretch` reads the stored state, so it returns the stretch that was just set even for an image that is not on screen. The same pattern holds for cuts, colormap and viewport.

This is asserted by `tests/test_image_viewer_logic_implementation.py::test_apply_hooks_gated_on_displayed_image`: it sets a `LogStretch` on an image that is no longer displayed, checks that no `_apply_*` hook fires, and then checks that `get_stretch(image_label="first")` still returns the `LogStretch`. The deferred stretch is applied when the image becomes displayed again (via the `_apply_*` sequence at the end of `load_image`).

One clarification on the diff: the gating itself is not new in this PR — it merged with the template-hooks work. These lines show as changed only because of the `value` → `stretch` parameter rename (#111).

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: The intent was to document two things that are implementation-specific rather than contract-level: this implementation validates against the *matplotlib* registry specifically, and it skips validation entirely when matplotlib is not installed. The contract (updated in this PR) only requires "raise `ValueError` for a name the backend does not support" — it does not mandate any particular registry, so that detail cannot live in the interface docstring.

But your push-back is fair on the execution. A docstring cannot be partially inherited, so documenting one implementation-specific note drags a near-verbatim copy of the interface's Parameters/Raises sections along with it — reintroducing exactly the docstring-drift risk the rest of this PR exists to fix. Worse, because backends inherit `ImageViewerLogic` docstrings through `docs_from_super_if_missing`, any backend that does not write its own `set_colormap` docstring would now claim to validate "against the matplotlib colormap registry" even if it validates against its own (Copilot's comment on the decorator brushes against the same problem).

The cleaner resolution: drop this override so `set_colormap` inherits the interface docstring via `@docs_from_interface`, and leave the implementation-specific behavior documented on `_validate_colormap_name` (which already carries the full matplotlib/optional-dependency story) and/or in the class docstring. No change has been made pending your call.

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Same question as on `set_colormap`, and the answer has the same shape. The implementation-specific content here is:

1. How a str/path is interpreted — FITS, or ASDF for a `.asdf` suffix. The interface does not mandate file formats, so this can only be documented on the implementation.
2. The Notes semantics: reloading an existing label keeps that label's cuts/stretch/colormap while resetting data, WCS and viewport. This was deliberately kept implementation-level — it is tested in `tests/test_image_viewer_logic_implementation.py::test_load_image_existing_label_keeps_its_settings`, not in the shipped compliance suite, because mandating it for every backend seemed too strong without discussion. The contract-level piece ("re-using an ``image_label`` replaces the image", #112) went into the interface docstring in this same PR.

The costs are also the same as on `set_colormap`: the Parameters section is a near-duplicate of the interface's (drift risk), and via `docs_from_super_if_missing` a backend without its own `load_image` docstring inherits the settings-retention promise even if it actually resets settings on reload.

Two ways to resolve it, your call:

- **Drop the override** and move both implementation-specific points into the class docstring (or the `_load_fits`/`_load_asdf` helpers). Keeps a single source of truth for the public API docs.
- **Promote settings-retention to contract level** — put it in the interface docstring and add a compliance test. That makes the override unnecessary for reason 2, but it is a real policy decision, not a docs fix.

No change made pending your decision.

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: It could be narrowed, but it should not be — this is the restore-and-reraise pattern, and the breadth is the point. The handler swallows nothing (it ends in a bare `raise`); its only job is to put the viewer back into a consistent state for *any* failure, because by the time the loaders run, `_displayed_image_labels` has been cleared and the label's entry replaced with a fresh `ViewportInfo`.

The set of exceptions the four load paths can raise is not realistically enumerable: `CCDData.read` raises `OSError` for malformed FITS but `ValueError` for a missing `BUNIT`; asdf raises its own validation errors; `_load_array` raises `ValueError`; and — the concrete case — the `os.PathLike` bug Copilot found in this same method raises `AttributeError`. Narrow the catch to, say, `(OSError, ValueError)` and that `AttributeError` escapes the handler after display tracking was cleared: the viewer permanently stops applying cuts/stretch/colormap to the image still on screen, and the half-initialized entry shadows the real one. The reproduction in the Copilot thread confirms the current code restores state correctly after exactly that unanticipated exception type.

(Also worth noting: the `try/except` predates this PR — it merged with the earlier load-image work; only the `file` → `data` rename touched these lines.)

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: It doesn't — `get_image` on this branch has no docstring of its own:

```python
def get_image(
    self,
    image_label: str | None = None,
    **kwargs,  # noqa: ARG002
) -> ArrayLike | NDData | CCDData:
    image_label = self._resolve_image_label(image_label)
    return self._images[image_label].data
```

It picks up the interface docstring through the `@docs_from_interface` class decorator, and this PR's only `get_image` change is *to* that interface docstring (the #112 fix: a real summary line and corrected parameter text).

This thread appears to be anchored on a stale hunk: it points at a removed line of the private `_catalog_conversion_wcs` helper's docstring, from the full-stack diff before the base PRs merged. That helper is not part of the interface (its docstring changed in the already-merged catalog-WCS work), so there is nothing being overridden there either. If the question was really aimed at the docstrings this PR *does* add on the implementation — `set_colormap` and `load_image` — those are answered in their own threads.

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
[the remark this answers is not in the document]
mwcraig: add it

pyproject.toml
mwcraig: Leave this as-is

src/astro_image_display_api/image_viewer_logic.py
mwcraig: leave this one as-is

src/astro_image_display_api/image_viewer_logic.py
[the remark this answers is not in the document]
mwcraig: How about renaming the decorator so that it is called docs_from_image_logic_if_missing or something like that? Nothing to implement on this one yet, just respond to the question.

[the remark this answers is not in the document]
mwcraig: A rename is a good idea, and right now is the cheapest it will ever be: the decorator has never been in a released version (0.2.1 predates the template-hooks merge), and the astrowidgets backends currently carry their own local copies of `docs_from_super_if_missing` rather than importing this one, so nothing downstream breaks.

Two ways to make the name truthful:

1. **Rename to match the behavior.** `docs_from_image_logic_if_missing` works; the fully unambiguous form is `docs_from_image_viewer_logic_if_missing` (matches the class name exactly, at the cost of being a mouthful). Either is fine.
2. **Fix the behavior to match the name.** Keep `docs_from_super_if_missing` but copy from the decorated class's actual MRO (`cls.__mro__[1:]`) instead of hard-coding `ImageViewerLogic`. For every current backend the result is identical (they subclass `ImageViewerLogic` directly), but it stays correct if a backend ever subclasses another backend, and it stops the decorator from silently doing the wrong thing if someone applies it to a class that is *not* an `ImageViewerLogic` subclass — today it would happily paste `ImageViewerLogic` docs onto an unrelated class.

Slight preference for option 2: it makes the decorator generic (same semantics as `inspect.getdoc`'s inheritance walk, applied eagerly) rather than tied to one class name, and the existing name becomes accurate for free.

One reason the decorator should stay in some form, in case "just delete it" comes up: Python's own tooling (`help()`, `inspect.getdoc`, Sphinx autodoc) already inherits missing docstrings through the MRO, but the shipped compliance test `test_every_method_attribute_has_docstring` checks raw `attr.__doc__`, which does not walk the MRO. A backend that overrides a public method without writing a docstring passes `help()` but fails the compliance suite — the decorator is what closes that gap.

Nothing implemented on this, per your note.

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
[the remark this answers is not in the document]
mwcraig: Done in 2115aa5: the str/PathLike branch is now `os.fsdecode(data).endswith(".asdf")`, with a regression test (`test_load_image_accepts_non_pathlib_pathlike`) that loads a FITS file through a `__fspath__`-only path class and checks `.asdf` routing on a bare `os.PathLike`. Full suite passes (90 tests).

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Reading the 👍 here (in contrast to the explicit "leave as-is" on `load_image` and the test extra) as agreement with the recommendation: the override is dropped in 2115aa5. `set_colormap` now inherits the interface docstring via `@docs_from_interface` — verified that `ImageViewerLogic.set_colormap.__doc__` matches the interface text and the compliance docstring test still passes. The matplotlib-registry/optional-validation specifics remain documented on `_validate_colormap_name`. If the 👍 meant something else, say so and it can be restored.

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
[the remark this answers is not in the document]
mwcraig: Go with 1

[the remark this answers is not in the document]
mwcraig: Done in 2048a59: renamed to `docs_from_image_logic_if_missing` (your suggested name), updated `__all__`, and the docstring summary line now says "from `ImageViewerLogic`" — which also resolves the original comment on this thread. Full suite still passes (90 tests).

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/api_test.py
```diff
@@ -83,6 +84,17 @@ def _get_catalog_labels_as_set(self):
...
+    @staticmethod
+    def _image_data_as_array(image_data):
+        """
+        Extract a plain array from whatever ``get_image`` returned.
+
+        The API does not specify the return type of ``get_image``, so
+        accept anything with a ``data`` attribute (`~astropy.nddata.NDData`
+        and friends) or an array.
+        """
+        return np.asarray(getattr(image_data, "data", image_data))
```
mwcraig, line 97: Isn't this imposing a requirement on the image data that is not part of the interface definition?

src/astro_image_display_api/api_test.py
```diff
@@ -983,30 +1008,30 @@ def test_catalog_conversion_uses_single_image_wcs(self, data):
         np.testing.assert_allclose(tab["x"], [10.0, 20.0], atol=1e-6)
         np.testing.assert_allclose(tab["y"], [30.0, 40.0], atol=1e-6)
 
-    def test_load_catalog_ambiguous_wcs_raises(self, data):
-        # Regression test for #91: with several images loaded there is no
-        # way to know which image's WCS to use to convert the catalog's
-        # sky coordinates to pixels, so loading must raise instead of
-        # silently using the last-loaded image's WCS.
+    def test_load_catalog_uses_displayed_image_wcs(self, data):
```
mwcraig: show me where the interface requires the behavior

src/astro_image_display_api/image_viewer_logic.py
```diff
@@ -814,3 +1060,26 @@ def get_viewport(
                 fov = viewport.fov
 
         return dict(center=center, fov=fov, image_label=image_label)
+
+
+def docs_from_image_logic_if_missing(cls):
```
mwcraig: Make this docs_from_image_viewer_logic_if_missing

src/astro_image_display_api/api_test.py
mwcraig: Strictly speaking, yes — and the helper is the smallest imposition that makes the #99 round-trip assertions testable at all. The interface documents `get_image` as returning the data of a loaded image but annotates it `Any` and promises nothing about the form. To check that loaded pixels survive a round trip (what #99 asked for), the suite has to get a numeric array out of *something*, so the helper accepts the two forms the ecosystem actually uses — a bare array, or an NDData-style object carrying it in `.data` — rather than demanding one exact type.

The gap you're pointing at is real, though: a backend that returned, say, a handle object with the pixels under a different attribute would satisfy the letter of the interface and still fail `test_load`/`test_get_image`. Two honest ways to close it:

1. **Tighten the interface** (recommended): add one sentence to `get_image`'s Returns — the returned object must be the loaded data, either array-like or exposing the array via a ``data`` attribute (`~astropy.nddata.NDData` and subclasses). That makes the round-trip test a documented requirement instead of an implicit one, and both existing backends already comply.
2. **Weaken the test**: drop the pixel comparison and only assert the label is registered. Then the shipped suite can no longer catch a backend that scrambles or drops data on load, which was the point of #99.

If you want option 1 it is a one-sentence interface docstring change; say the word and it goes in with the decorator rename.

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/api_test.py
mwcraig: It doesn't — you've found a real contract/test mismatch. The interface's `load_catalog` Raises section still documents the *old* behavior:

> Also raised if a pixel/sky conversion is required but the choice of image WCS is ambiguous because several images are loaded.

Nothing in the interface says the displayed image's WCS breaks the tie; worse, "the displayed image" is not an interface concept at all — no method in the protocol exposes which image is displayed, so a backend that (for example) shows several loaded images side by side has no well-defined way to satisfy this test.

How it got here: when the catalog-WCS work changed the reference implementation to prefer the single displayed image's WCS, `test_load_catalog_ambiguous_wcs_raises` started failing against the reference implementation, so it was rewritten into `test_load_catalog_uses_displayed_image_wcs` to keep the suite green (this happened in the already-merged branch; this PR's diff only shows it because of the stale full-stack view). That was the wrong fix direction: the shipped compliance suite ended up enforcing reference-implementation behavior the contract never states, while the contract text it *does* state (raise on ambiguity) is no longer tested as written.

Three ways to reconcile, in rough order of preference:

1. **Move the displayed-image tests out of the shipped suite** — `test_load_catalog_uses_displayed_image_wcs` and the displayed-WCS expectation in `test_load_catalog_pixel_only_with_multiple_images` belong in `tests/test_image_viewer_logic_implementation.py` with the other displayed-image behavior. Soften the interface's Raises wording to "if the backend cannot determine which image's WCS to use", so a backend that raises on any multi-image ambiguity and one that disambiguates are both conformant, and have the shipped suite test only the single-image and no-WCS cases it can state contract-language for.
2. **Promote the behavior to contract level** — define "displayed image" in the interface and document the tie-break in `load_catalog`. That is a much bigger API decision (it adds a whole concept to the protocol) and probably wants its own issue, not a drive-by in this PR.
3. **Leave it** — not really defensible once stated out loud: the suite would keep failing backends for not implementing undocumented behavior.

No code pushed per your instruction; happy to implement option 1 (or whichever you pick) alongside the decorator rename.

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Will do — `docs_from_image_viewer_logic_if_missing` it is. Not pushed yet per your instruction to respond only; it will go in the next commit, along with whatever you decide on the two threads above (the `get_image` Returns tightening and the displayed-image-WCS test placement).

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/api_test.py
mwcraig: Let's go with 1.

src/astro_image_display_api/api_test.py
mwcraig: Do 1

src/astro_image_display_api/api_test.py
mwcraig: Done in 3808023: `get_image`'s Returns section now requires the returned object to be the data loaded for that label, either array-like or exposing the array via a `data` attribute (as `NDData` and subclasses do) — replacing the old "exact type is not specified" wording. The `_image_data_as_array` helper's docstring now cites that contract instead of apologizing for the lack of one.

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/api_test.py
mwcraig: Done in 3808023: `test_load_catalog_uses_displayed_image_wcs` is deleted from the shipped compliance suite, and `test_load_catalog_pixel_only_with_multiple_images` now only requires that a pixel-only catalog load succeed with several images (x/y round-trip); which WCS, if any, fills in the sky column is backend-dependent. The displayed-image-WCS coverage moved to `tests/test_image_viewer_logic_implementation.py` as reference-implementation tests, and `load_catalog`'s Raises wording is softened to "the backend cannot determine which image's WCS to use (e.g. because several images are loaded)". Suite is 91 passing (−1 moved, +2 implementation tests).

*This comment was written by Claude at Matt Craig's direction.*

src/astro_image_display_api/image_viewer_logic.py
mwcraig: Done in 3808023: `docs_from_image_viewer_logic_if_missing`, in the `def` and `__all__` (the only two references in the repo).

*This comment was written by Claude at Matt Craig's direction.*

